PR #45, opened 2019-06-07
add rustic-cargo-current-test and bind it to "C-c C-c C-c"

### LINKED ISSUE #43: Add rustic-cargo-test-current
Hi!
It would be nice to have function similar to one in [cargo](https://github.com/kwrooijen/cargo.el) package:
```
C-c C-c C-f | cargo-process-current-test.
```
It runs test near the point, which is very useful during development when you want to run one specific test.

TatriX:
This new command can be used to run only the test near the point.
It should work for all the cases bellow:
```
fn test_nothing() {
    assert!(false);
}

fn test_something() {
    assert!(true);
}

mod tests {
    #[test]
    fn wut_nothing() {
        assert!(false);
    }

    #[test]
    fn wut_something() {
        assert!(true);
    }
}
```

Tests will be added in the following commit.

Closes #43.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -111,6 +111,47 @@ If ARG is not nil, use value as argument and store it in `rustic-test-arguments'
                               :process proc
                               :mode mode)))
 
+;;;###autoload
+(defun rustic-cargo-current-test ()
+  "Run 'cargo test' for the test near point."
+  (interactive)
+  (rustic-compilation-process-live)
+  (rustic-compilation-start (list rustic-cargo-bin "test" (rustic-cargo--get-current-fn-fullname))
+                            :buffer rustic-test-buffer-name
+                            :process rustic-test-process-name
+                            :mode 'rustic-cargo-test-mode))
+
+(defun rustic-cargo--get-current-fn-fullname()
+  "Return full name of the fn around point including module name if any."
+  (let ((mod (rustic-cargo--get-current-mod))
+        (fn (rustic-cargo--get-current-fn-name)))
+    (if mod
+        (concat mod "::" fn)
+      fn)))
+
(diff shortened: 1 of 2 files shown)


rustic-cargo.el
```diff
@@ -111,6 +111,47 @@ If ARG is not nil, use value as argument and store it in `rustic-test-arguments'
...
+
+(defun rustic-cargo--get-current-mod ()
+  "Return mod name around pount or nil."
+  (save-excursion
+    (when (search-backward-regexp rustic-cargo-mod-regexp nil t)
+      (match-string 1))))
+
+(defun rustic-cargo--get-current-line-fn-name()
+  "Return fn name from the current line or nil."
+  (save-excursion)
```
brotzeit: The argument is missing.

TatriX: What do you mean?

brotzeit: You didn't add the function body to the `save-excursion` call. This will throw an error.

TatriX: Ah, right, thanks ;) I wish it had type system around. Btw, have you seen [Elsa](https://github.com/emacs-elsa/Elsa)?

brotzeit: Yeah, but I didn't try it. You can run `eval-buffer` to catch those errors. Or `byte-recompile-directory` should also work.

TatriX: Well, it's not an error from the elisp perspective, so those commands won't help in this case.

TatriX:
I'm not sure why build is failing. It looks like something just timed out. Could you please restart it?

rustic-cargo.el
brotzeit: Oh, you just helped me to find a bug in remacs. Thanks ;)

TatriX:
I restarted it by force pushing with white-space cleanup change and it works now ;)

PR #47, opened 2019-06-14
Fix playground url

TatriX:
Currently it's pointing to `play.rustic-lang` which should presumably be `play.rust-lang`.


brotzeit:
It seems I did some refactoring after I tested this function. Thanks for catching this!

PR #54, opened 2019-07-23
Add open/build documentation option and check for cargo outdated

sehnsucht13:
- Add option to open documentation/build it as well or to simply build documentation without opening it.
- Add open documentation option to popup menu
- Add an extra check for cargo-outdate. If the binary is not installed in path, rustic seems to start installing an unrelated package for me. The new check will simply display a warning and ask the user to install it.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -427,13 +427,16 @@ If BIN is not nil, create a binary application, otherwise a library."
 (defun rustic-cargo-build-docs ()
   "Build the documentation for the current project."
   (interactive)
-  (rustic-run-cargo-command "cargo doc"))
+  (if (y-or-n-p "Create documentation for dependencies?")
+      (rustic-run-cargo-command "cargo doc")
+    (rustic-run-cargo-command "cargo doc --no-deps"))
+  )
 
 ;;;###autoload
 (defun rustic-cargo-open-docs ()
   "Open the documentation for the current project in a browser."
   (interactive)
-  (rustic-run-cargo-command "cargo doc --open"))
+  (rustic-run-cargo-command "cargo doc --open --no-deps"))
 
 (provide 'rustic-cargo)
 ;;; rustic-cargo.el ends here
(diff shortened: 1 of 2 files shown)


rustic-cargo.el
```diff
@@ -380,6 +388,7 @@ If BIN is not nil, create a binary application, otherwise a library."
 ;; Interactive
 
 (defun rustic-run-cargo-command (command)
+  "Run the specifiend COMMAND with cargo."
```
brotzeit: typo

brotzeit:
Maybe we should offer to install it ?

rustic-cargo.el
sehnsucht13: Thanks!

sehnsucht13:
I have thought about adding an install command. The only thing that I am not sure about are install permissions. Would calling `cargo install cargo-outdated` require any type of permissions(sudo, admin permission on windows...) or are those not needed? I just did not want it to fail silently on the user if the option is included. 

On another note, do you know why the build failed previously? Is it a problem with the spinner package or did I somehow introduce something?

Also, would you like me to squish the commits into one?

sehnsucht13:
Update to this, I asked on the Rust Discord and it does not seem like you would need any time of sudo/administrator permissions ever so the install command cannot really fail for that reason. I will take care of adding an auto-install option tommorow

sehnsucht13:
Done, I did not notice that there was a `rustic-cargo-install-crate-p` command. I added two optional arguments to that called `ymsg` and `nmsg`. The first is displayed if the user decides they want to install the crate. The second is used if they do not want rustic to install the crate.

The idea behind these is that in the case of installing cargo-outdated(user saying yes) then a message will show which tells them to run `rustic-cargo-outdated` again after the install is complete. If they choose no then there is a warning message which tells the user that `cargo-outdated` must be installed in order to display the information.

sehnsucht13:
I am not too sure as to why the build is failing.

brotzeit:
Yeah, I know this issue with the tests. This seems to be a problem for many projects. It looks like a timeout...

I will take a look at it.

brotzeit:
Sorry for the delay. I will review this today.

sehnsucht13:
Its no problem. I have been quite busy myself. Take your time!

Issue #55, opened 2019-07-23
New rustic-popup proposal


sehnsucht13:
Would you be interested in adding a transient interface for cargo and replace the current popup with it? It will be similar to the interface that magit has. This will let us define more options/commands/flags in an easier manner and the work in maintaining the current popup can be removed by relying on transient. If you are interested, I can make a PR with a proof of concept(a subset of commands).

brotzeit:
That would be great. I think I've tried to implement it this way, but there was an issue I don't remember =)

But I'm sure whatever it was, it can be solved.

sehnsucht13:
Ok, I will start implementing something small which will cover what is in the popup right now and then if accepted, I can expand.

tarsius:
If one of you has a half finished version still laying around, then I could possibly help with that.

brotzeit:
close in favor of  #303

PR #56, opened 2019-07-28
Preview of the transient

sehnsucht13:
As mentioned in #55, I have started working on a transient for cargo. It is not finished yet but if you would like to, you can take a look at it/try to run it. All suggestions about layout/documentation/functions are welcome.

As of right now, it does not invoke any commands. I have simply created a function which splits the arguments and prints them in the minibuffer. The most complete transient is the `doc` one which only needs to have calls added for cargo to run the command.

You can run the main transient with the command `rustic-transient-popup` after you evaluate the `rustic-transient.el` file.

I have split the menu into two parts. There are the quick shortcuts(current rustic-popup commands) and there is the full menu(right side) which gives complete control to the flags passed by cargo.

Here are some images:

transientMain

transientDoc

### DIFF
--- a/rustic-transient.el
+++ b/rustic-transient.el
@@ -1,17 +1,8 @@
-;;; rustic-transient.el --- Transient based popup-*-lexical-binding: t-*-
+;;; rustic-transient.el --- Transient based popup -*-lexical-binding: t-*-
 
 ;;; Code:
 
 (require 'rustic-cargo)
-
-;;;###autoload
-(define-infix-command rustic-transient:-trace ()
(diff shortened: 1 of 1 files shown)


brotzeit:
Nice! I will give it a try when you say it's finished.

brotzeit:
We also need to update the readme.

sehnsucht13:
Will do. I will need some time to finish it and make sure that everything works properly. I will close this pull request for now so that circleci does not start testing it every time i push. This was more of a heads up on progress.

PR #58, opened 2019-09-07
use right keymap name in rustic-mode docstring

aaptel:
rustic defines a rustic-mode-map keymap, not rustic-map. this fix will
properly show the available bindings when running M-x
describe-mode (C-h m) while in rust source buffer.


brotzeit:
Thanks!

brotzeit:
The build isn't actually failing. This happens from time to time...

PR #60, opened 2019-09-08
rustic-racer.el: use right var name in error msgs

aaptel:
racer-rust-src-path doesn't exist, replace it with
rustic-racer-rust-src-path.


brotzeit:
Thanks!

PR #68, opened 2019-10-01
Expired Elpa Keys

dickmao:
cf melpa/melpa#6462

The elpa key expiration has wrought havoc among elisp repos that actually conduct proper testing, proving once again that no good deed goes unpunished.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,24 @@ language: emacs-lisp
 
 sudo: required
 
-dist: trusty
-
 notifications:
   email: false
 
 cache:
   - directories:
-      - "$HOME/emacs/"
+      - $HOME/emacs
+      - $HOME/.cache/pip
+      - $HOME/.emacs.d
+      - $HOME/.cargo
+      - $TRAVIS_BUILD_DIR/.cask
 
 matrix:
+  allow_failures:
+    - env: EMACS_VERSION=snapshot
   fast_finish: true
-  
-env:
-  - EMACS_VERSION=26.1
-  - EMACS_VERSION=snapshot
+  include:
+    - env: EMACS_VERSION=26.1
+    - env: EMACS_VERSION=snapshot
 
 before_install:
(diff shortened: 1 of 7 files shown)


brotzeit:
Wow, thanks! I will take a look at it and try to fix the tests.

dickmao:
The failures are due to my removing `cargo install --force rustfmt-nightly`.  That particular invocation seems compile the universe all over again (taking about 12 minutes).

I don't understand the rust toolchain enough to do this correctly.  Today was the first time I installed rust on my laptop.

brotzeit:
Nice!

dickmao:
> `all` doesn't do what it used to do: generate the `*.elc` files.

In #97, I restore that behavior.  To explain, I like `make` merely to verify the code compiles (ideally without warnings, which this particular project cannot vouch for).  Having `make` leave behind `.elc` files seems unnecessary and potentially dangerous.  The unintended consequences of stale byte-compiled elc files are widely felt by unsuspecting emacs devotees.

phillord:
> Also why is the `EMACS=$(EMACS)` thing needed?

I use that trick a lot. You can reset `EMACS` to be what ever you need this way without changing the make file.

For instance, m-buffer uses this.

https://github.com/phillord/m-buffer-el/blob/master/Makefile

dickmao:
Well, in my case it wasn't strictly needed (I suspect I copied it from somewhere else).  I had fixed it in #97 .

If the env variable EMACS is exogenously set outside make, then it naturally flows through to cask.  If it's not set, cask assumes EMACS is unversioned "emacs".  So in both cases, explicit assignment of `EMACS=$(EMACS)` is redundant.

You might have a recipe command that isn't cask which requires the env variable EMACS to be set, so such an assignment may have its uses.  But I see this thing almost always with cask, and it's superfluous.

phillord:
Yep, I normally use a "makefile-local" which I include and set my defaults up there, then conditionally include this. Allows me to set defaults without inflicting them on anyone else.

PR #80, opened 2019-12-01
make rustic-compile match its docstring

nikomatsakis:
The docstring claims that rustic-compile, if invoked without
arguments, will prompt for the command. But in fact it did not do so.

### DIFF
--- a/rustic-compile.el
+++ b/rustic-compile.el
@@ -439,7 +439,7 @@ Otherwise use provided argument ARG and store it in
 `compilation-arguments'."
   (interactive "P")
   (let* ((command (setq compilation-arguments
-                        (if arg
+                        (if (not arg)
                             (read-from-minibuffer "Compile command: ")
                           rustic-compile-command)))
          (dir (setq compilation-directory (rustic-buffer-workspace))))


nikomatsakis:
@brotzeit the errors look spurious -- that said, even with this change, when I execute `rustic-compile`, it still doesn't ask me for a command to run. I would assume this is meant to act analogously with `M-x compile` and `M-x recompile`, no? Maybe the argument should just be required?

brotzeit:
Sorry, I've been busy. I'm a little confused, since `compile` shouldn't be asking for a command by default, or maybe I made a customization that I'm currently not aware of ;)
AFAIK `compile` is only intended to prompt you for a command when you provide `universal-argument`. So when you run `C-u` `M-x` `compile` you will get asked for a command.

But I will take a look at it if I'm missing something here and `compile` behaves differently that `rustic-compile`.

brotzeit:
You can use this function instead if you wan't to get asked by default:

```elisp
(defun my-rustic-compile ()
  (interactive)
    (let ((current-prefix-arg 4)) ;; emulate C-u
    (call-interactively 'rustic-compile)))
```

nikomatsakis:
@brotzeit What then is the point of `rustic-recompile`?

From what I can tell, `M-x compile` always prompts for a command and `M-x recompile` just re-executes the previous command. I would have expected the same behavior from rustic-compile and rustic-recompile.

I did have a wrapper similar to the one you described, but I figured this was likely a bug in rustic. 

But it's entirely possible I'm confused about something!

brotzeit:
`rustic-compile` is supposed to use the value of `rustic-compile-command` when called without `C-u`.
`rustic-recompile` makes use of the command that is stored in `compilation-arguments`.

So the point of `rustic-recompile` is to provide a way to easily recompile with a command that is not equal to your `rustic-compile-command`.

brotzeit:
I checked it again. Apparently I changed the default value of `compilation-read-command` sorry =)
I will change it so it works analogously to `compile`. Thanks for opening an issue!

brotzeit:
Fixed. Please reopen in case it still doesn't work.

PR #87, opened 2020-01-26
Refactor package.el dependency out of lsp client setup

hlissner:
This is a simpler way to determine if lsp-mode or eglot is available, without depending on package.el.

`(require X nil t)` returns nil if it fails to load `X`, but will return non-nil if it succeeds.


brotzeit:
Thanks!

PR #89, opened 2020-01-29
Remove extra paren

strangelittlemonkey:

PR #90, opened 2020-01-30
Fix prematurely terminated cond

hlissner:
Sorry about that. It seems I had one ) too many in #87, which wasn't correctly fixed in #89, causing `void-function t` errors. This PR fixes that.


brotzeit:
No problem, I also didn't take the time to review it ;)

Issue #85, opened 2020-01-14
Rustic lsp is still picking up rls even though its set to rust-analyzer


CSRaghunandan:
This is my configuration for rust: https://github.com/CSRaghunandan/.emacs.d/blob/master/setup-files/setup-rust.el

I use `exec-path-from-shell` as well. Upon inspecting I can see that the value of `rustic-lsp-server` is set to `rust-analyzer` and  the right path is set for the server command too.

But when I open any project, it runs `rls` instead of `rust-analyzer`. I tried setting the path of `lsp-rust-analyzer-server-command`, but that did not help too.

EDIT: found a way to fix this issue. I set `rustic-lsp-server` to `nil` and enabled `lsp` manually in a hook and that seems to have solved the issue for now. Maybe there was some changes to `lsp-rust` which caused this to break.

brotzeit:
Thanks for opening an issue. I definitely have to make some changes here. Is it possible that lsp-mode isn't required before rustic ?

CSRaghunandan:
> Thanks for opening an issue. I definitely have to make some changes here. Is it possible that lsp-mode isn't required before rustic ?

Yes, there is no need to run `(lsp)` to start `lsp-mode`. But right now, I'm running it manually while disabling rustic's handling of `lsp-mode`. 

Not sure what is causing `rustic` to pick up `rls` even though all the variables are set for `rust-analyzer`.

brotzeit:
The server priority also has to be changed. I added a test https://github.com/brotzeit/rustic/commit/1fc6c56d21cc60b287351ff0ec6f707a7f81cfed

CSRaghunandan:
Latest commit broke rustic for me.

As soon as I open a rust file, I get an error:

```
Debugger entered--Lisp error: (wrong-type-argument lsp--client nil)
  signal(wrong-type-argument (lsp--client nil))
  rustic-lsp-set-server(rust-analyzer)
  rustic-setup-lsp()
  rustic-mode()
  set-auto-mode-0(rustic-mode nil)
  set-auto-mode()
  normal-mode(t)
```

brotzeit:
lsp-mode seems to require lsp-clients not until you open a rust file. I added a check for the server priorities that gets called before `lsp`. Thanks for the fast response.

nmccarty:
Running on the latest melpa version (20200213.1139) is still giving me the same wrong-type-argument error. 
```
[...]

PR #94, opened 2020-02-17
Manually load lsp-rust

nmccarty:
This acts as a workaround for #85.
It appears lsp-mode has changed the way that loading lsp-clients works, and simply requiring
lsp-clients does not appear to be enough anymore, so we additionally directly require lsp-rust.


brotzeit:
It's odd that it works for me with requiring lsp-clients. But if this works for you. Thanks!

Issue #92, opened 2020-02-13
Please document reason for fork


tarsius:
I am just getting started with Rust and am trying to determine whether to use `rust-mode` or `rustic`. From the look of it, it will take quite some time to make an informed choice so I decided to ask you for some clarification to save myself and others in the same position part of the effort.

Your readme mentions some features that `rustic` has but `rust-mode` lacks. That's a nice start but many questions come immediately to mind that are not addressed by a simple list of features, such as:

1. Does `rust-mode` still lack those features?
2. Do you port `rust-mode` improvements to `rustic`? Other way around?
3. Why fork in the first place? Conflict?
4. Assuming the fork is due to them wanting to continue to provide just the major-mode and you wanting to do a batteries-including ide: why did you fork `rust-mode.el`? Couldn't you have contributed to that in the upstream repository and make `rustic` depend on that?
6. Would it make sense to break up `rustic` into multiple smaller packages? I had to install quite a few dependencies that I am not actually going to use just so I could compile.
5. What happens if both packages are installed?

Errr.... I just saw that you commented on the issue tracker of the **`rust-mode`** repository.
That leaves me more confused than before, but also more hopeful.

tarsius:
Re https://github.com/brotzeit/rustic/pull/91.
Re https://github.com/rust-lang/rust-mode/pull/353.
:stuck_out_tongue_winking_eye:

brotzeit:
I'm at work now, but I will answer your questions when I'm home and add it to the readme.

brotzeit:
I will provide some background to this situation.

About two years ago I started adding some features that I wanted to be in rust-mode(https://github.com/rust-lang/rust-mode/issues/255). Unfortunately rust-mode doesn't have active maintainers for quite a while so it was difficult to get pull requests merged, that were bigger than some lines of code. At some point I was kind of frustrated and somebody, I think it was actually you @tarsius, told me he took over maintainership for a package(I really think it was you, so magit^^) because there was no maintainer.

So I decided to give it a try and created a pull request on melpa. Half a year ago I suggested to also maintain rust-mode (https://github.com/rust-lang/rust-mode/issues/317). The result was that @mookid and I decided to give it a try. I leave out the details because you can read it in the issue.

We discussed how to proceed, via mail. The idea was what you actually mentioned in 4). Due to a lack of time we didn't make those changes.

[...]

PR #95, opened 2020-02-17
Major cleanup (whitespace, headings, warnings, circular deps)

tarsius:
As promised in #92, here comes a major cleanup.

1. Cleanup whitespace (same as #91 but now for all libraries).
2. Use outline-minor-mode compatible section headings. 415e619ed4bdffe0c50d9e49b8b8676692def718
3. Address all byte-compilation warnings (approximately 60!).
4. Break circular dependencies between libraries. 6711095d4ce6ee876bf4af0c13aecb0188eb7ce3

Please make sure the fixes in 1c75e74ed6b29d18ca18471275fbdcee7081f6ee are correct and don't just further hide bugs.

Next I would like to make all libraries except `rustic.el` itself optional.  That basically works already, but needs more testing.


brotzeit:
Nice, thanks! I want to review this properly so I maybe won't merge this before the next weekend.

tarsius:
If you checkout this branch and in the `rust-mode` repository `tarsius/backport` you now can get a human readable comparison between the two modes like so:

```shell
cd rust-mode
cp ../rustic/rustic.el rust-mode.el
sed rust-mode.el -i -e s:rustic:rust:g
```

Then view the diff in Magit.

rustic-compile.el
```diff
@@ -478,5 +478,6 @@ Otherwise use provided argument ARG and store it in
     (rustic-compilation-process-live)
     (rustic-compilation (split-string command) :directory dir)))
 
+;;; _
```
brotzeit: Out of curiosity, what is the purpose of this ?

brotzeit:
It seems the user needs to change his config in order to use all rust components ? IMO it should be optional to disable the other features, not to enable them ?

Now emacs complains about void variable `rustic-rls-pkg` when opening a rust file.

tarsius:
> It seems the user needs to change his config in order to use all rust components ? IMO it should be optional to disable the other features, not to enable them ?

I have removed those changes for the time being and moved them to the `lighttweight` branch. Also fixed some issue.

(Another issue that I ran to, even on `master`, is that when one has `lsp-mode` installed without configuring it, then an error occurs in `rustic-lsp-mode-setup` because `rustic-lsp-server` isn't in `lsp-client`. I don't know how that is supposed to get in there. So I had to disable `rustic-lsp-setup-p`.)

> Now emacs complains about void variable `rustic-rls-pkg` when opening a rust file.

I have fixed that.

----

After splitting out the difficult parts to another branch, this should be safe to merge again.

brotzeit:
I can't push to your branch so I added the branch to this repo. I just moved some code to make things work again. If it is ok for you I would merge this when tests go green ?

tarsius:
> I can't push to your branch so I added the branch to this repo.

Hm, I'll have to look into how that happened.

> I just moved some code to make things work again. If it is ok for you I would merge this when tests go green ?

Give me a moment to look into that.

tarsius:
Okay that's one way of doing it (in one of the commits that I am holding back for now I instead moved both `rustic-re-item-def-imenu` and ` rustic-imenu-generic-expression` right before `rustic-mode-syntax-table` instead. If you move them to eof, then you should should at least *declare* somwhere before `rustic-mode-map`.

But I would rather finish the remaining two commits before anything is merged. I just needed a break; it's rather boring work.

brotzeit:
Yeah I know, that's why I'm especially grateful ;)
I'm on my way home, if you want you can fix it now or wait.

tarsius:
I would like to let it rest until Monday and if you in turn can only work on it on the week end, then we should probably delay until the end of *next* week.

brotzeit:
I was a little busy this week and wanted to take the time to Review it. Monday sounds good.
Have a nice weekend.

tarsius:
Do you run the tests locally? It takes forever on Travis.
Do you have setup instructions? I don't just want to blindly copy the Travis configuration (especially considering that what it does seems to be quite expensive).

tarsius:
I am currently doing "testing by using". I also looked at the test output, but I cannot wait > an hour after every little change. I noticed it complained a lot about some functions being void.

We no longer explicitly load the libraries that define that, so that's not a complete surprise, but they should be autoloaded, i.e. they should get loaded when first used. I think this does not work properly due to a bug in `.travis.yml`, but I have very little experience with Cask, so I am not sure.

Shouldn't `cask build` also be run? Why does doing that not generate `rustic-autoloads.el` (as I just found out by doing it locally)? How does one generate that file when using Cask? Does `cask exec ert-runner` load `rustic-autoloads.el` if it is present?

Or in other words, it appears that the test setup also needs some work (and I am not the one to do that).

Meanwhile we might have to give up on not explicitly requiring all additional libraries until we can reliably and somewhat quickly test whether that actually works as expected.

tarsius:
> Do you have setup instructions?

Oh! You do. But they appear to be extremely incomplete and result in one failed test and all other tests being "unknown", whatever that means.

brotzeit:
> I am currently doing "testing by using". I also looked at the test output, but I cannot wait > an hour after every little change. I noticed it complained a lot about some functions being void.

I wouldn't wait for travis, but run the tests locally. It takes some time to download the rust lsp server, but I had several issues regarding rust-analyzer that does not start, so I decided to add this test.

> We no longer explicitly load the libraries that define that, so that's not a complete surprise, but they should be autoloaded, i.e. they should get loaded when first used. I think this does not work properly due to a bug in .travis.yml, but I have very little experience with Cask, so I am not sure.

I'm actually still inclined to keep providing all features, since most users want that(I would have no problem with doing it the other way round, but I'm sticking with the majority of users).

> Shouldn't cask build also be run? Why does doing that not generate rustic-autoloads.el (as I just found out by doing it locally)? How does one generate that file when using Cask? Does cask exec ert-runner load rustic-autoloads.el if it is present?

I don't really know, but what I do know is that it works this way ;)
I don't know what's the purpose of this autoloads file. I don't have this on my disk. Maybe it's a configuration thing ?

> Or in other words, it appears that the test setup also needs some work (and I am not the one to do that).

Sure it can always be done better, but it works and it saved me a lot of nerves. Look at master, it's green. I have been fiddling with it on another branch so I know it works, currently. The issues you've seen seem to come up every now and then :D
Most of the times it works again when I clear the cache, don't ask me why...

> Meanwhile we might have to give up on not explicitly requiring all additional libraries until we can reliably and somewhat quickly test whether that actually works as expected.

As mentioned before, I would keep requiring all libraries and add some tests to check whether it works with rustic.el only. Sorry, but I already see users opening issues why the features they were using are not available anymore.

That being said I really appreciate your work here and I can also finish your work if you want. I know that you have tons of other things you want to improve ;)

tarsius:
The idea isn't to force users to explicitly opt-in to functionality that was previously available out of the box. You made it clear that you don't want that and I accept that.

What I am trying to do right now is to merely to avoid loading libraries until the things that they define are actually being used. This libraries should automatically be loaded because of a technique called "autoloading". This means that a special file `rustic-autoloads.el` is generated from all the functions and variables whose definitions are prefixed with `;;;###autoload`. Once such a symbol is actually used the library that contains the real definition is loaded.

Now there are a few issues with that: It appears Cask does not generate that, and I am having difficulties running the tests.

> As mentioned before, I would keep requiring all libraries and add some tests to check whether it works with rustic.el only.

I'll look into that.

tarsius:
Travis seems to refuse to test the latest tip of my pr, see https://github.com/brotzeit/rustic/pull/95/checks. Could you tell it please run the tests? I made some experimental changes that are specifically intended to test how travis tests, so it is a bit frustrating that it is right now deciding to go on strike. :roll_eyes:

brotzeit:
> I made some experimental changes that are specifically intended to test how travis tests, so it is a bit frustrating that it is right now deciding to go on strike.

Yeah, I know what you're talking about.
I have to do some more work, but hopefully I can take a look at it later but probably it will be tomorrow. Maybe it's better for you to work on something else for now ;)

brotzeit:
Wow, looks great! I would merge this now ? What about #97 ?

tarsius:
Well obviously I still have to cleanup the various `wip` commits, but I could do that now.

I wanted to play with `emacs-module-rs` instead, but got stuck on the first step. https://github.com/ubolonton/emacs-module-rs/issues/25 So I got some time.

brotzeit:
Ok, tell me when you're ready.

tarsius:
I think we are good to go. (Once the tests complete.)

brotzeit:
:ok_hand:

brotzeit:
Finally "rebase and merge" is set so I can't forget it again :sweat_smile:

Thanks again. The next PR will be merged properly ;)

PR #97, opened 2020-02-24
restore build's littering of elc files

dickmao:
Spruce up Makefile.

Let `make` create elc files as it once did.


brotzeit:
Thanks! I will wait for @tarsius, he put much effort into his PR. Maybe we can combine them ?

PR #103, opened 2020-04-26
In emacs >= 26.2, use replace-buffer-contents after formatting

tspiteri:
This saves the position and the markers in the buffer, so there is no need to go to` rustic-save-pos`. This plays better with e.g. grep, as otherwise all grep markers would be destroyed when replacing the contents of the buffer with `copy-to-buffer`.

While `replace-buffer-contents` was added in Emacs 26.1, it was broken for non-ASCII contents and would corrupt the buffer contents. This was fixed in 26.2, so we require version 26.2 for this.


brotzeit:
Thanks! Looks good.

PR #104, opened 2020-04-26
Fix typos

tarsius:


brotzeit:
Thanks!

tarsius:
Hm... I seem to be stuck in some alternative timeline.

PR #108, opened 2020-05-15
upstream: Re-implement rust-in-macro for performance

### LINKED ISSUE #107: Suspected poor performance from rustic-syntax-propertize
Hello! I apologize if I'm confused or misguided here, but I believe that the `rustic-syntax-propertize` function is locking up emacs frequently when editing any rust source file of non-trivial size. It seems that with anything larger than about 100 lines emacs becomes pretty much unusable.

This suddenly started happening a few weeks ago, and at first I thought it was my workstation, but then had the same thing happen after switching to another workstation, but only after upgrading my doom emacs installation. Unfortunately, I don't recall whether I updated rust/rls at the same time. I'm not proficient with emacs lisp, so from the little of the source I've looked at, it's unclear how much this function relies on rls for any functionality.

I tracked down this function as the culprit by running the profiler in emacs; as you can see in the screenshot, it's consuming ~75% of CPU and 55% memory during my sample run; that's pretty typical over the 2 machines and several samples I've run. Specifically, `rustic-ordinary-lt-gt-p` seems to be the problem function.

Screenshot from 2020-05-11 14-45-12

As I said before, this very much outside my area of expertise, so I might be completely wrong, but I'm hitting dead-ends on trying to find any other issue in other projects that explain it, so I thought I'd bring it up here. I _am_ running nightly rust, but I've been updating frequently since hitting this issue and the problem persists, so if that's the cause it's something more than a whoops commit.

Setting `syntax-propertize-function` to `nil` appears to completely resolve the problem, but then you don't get formatting while editing. That's been my workaround today since tracking this down.

brotzeit:
Thanks @phillord

@brownjohnf can you test this please.

close #107

### DIFF
--- a/rustic.el
+++ b/rustic.el
@@ -453,8 +453,71 @@ symbols."
   (let ((beg-of-symbol (save-excursion (forward-thing 'symbol -1) (point))))
     (looking-back rustic-re-ident beg-of-symbol)))
 
+(defvar-local rustic-macro-scopes nil
+  "Cache for the scopes calculated by `rustic-macro-scope'.
+This variable can be `let' bound directly or indirectly around
+`rustic-macro-scope' as an optimization but should not be otherwise
+set.")
+
+(defun rustic-macro-scope (start end)
+  "Return the scope of macros in the buffer.
+The return value is a list of (START END) positions in the
+buffer.
+If set START and END are optimizations which limit the return
+value to scopes which are approximately with this range."
+  (save-excursion
+    ;; need to special case macro_rules which has unique syntax
+    (let ((scope nil)
+          (start (or start (point-min)))
+          (end (or end (point-max))))
+      (goto-char start)
+      ;; if there is a start move back to the previous top level,
+      ;; as any macros before that must have closed by this time.
+      (let ((top (syntax-ppss-toplevel-pos (syntax-ppss))))
+        (when top
+          (goto-char top)))
+      (while
+          (and
+           ;; The movement below may have moved us passed end, in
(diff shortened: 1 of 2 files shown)


brownjohnf:
Yep, will do. Probably this weekend.

brotzeit:
I merge this now, but it would be nice if you could post the differences with these changes.

brownjohnf:
This works amazingly well! Including a screenshot of the profiler now, showing basically no resource usage by this code. Thank you SO much to both of you for troubleshooting/making this fix so quickly and helping out an ignorant emacs newb like myself :-).

@brotzeit I'd be happy to make a PR to doom emacs (how I'm consuming it) bumping the version to current master if you think it's ready to go.

Screenshot from 2020-05-16 07-54-31

brotzeit:
@brownjohnf Nice! And yeah that would be nice, I was planning to open a PR myself.

brownjohnf:
So after using this more, it definitely is better, but still not as snappy as I'd hope, or as fast as it seemed before the regression. I'm not sure what more to do about it though, and it'd be odd if these fixes make it fast enough in rust-mode but not rustic. Maybe I just have a slow machine.

brotzeit:
Hmmm, do you know that it's faster in rust-mode ?

brotzeit:
When I have the time, I will take a look at it, but I doubt I'm able to really improve it. I'm currently waiting for the native-comp branch in emacs to be more stable. I guess this will bring more performance improvements.

PR #109, opened 2020-05-19
Cache macro locations in rustic-syntax-propertize

phillord:
rustic-macro-scopes is not correctly cached during
rustic-syntax-propertize, meaning multiple calls are made to
rustic-macro-scope during syntax analysis.


brotzeit:
Thanks!

PR #111, opened 2020-05-21
upstream: Check for -> and => early in rust-ordinary-lt-gt-p

phillord:
Thanks to tspiteri

As discussed here https://github.com/brotzeit/rustic/issues/107, this should also speed up syntax-propertize for files with lots of `->` or `=>` symbols in it.


brotzeit:
Let's give it a try!

phillord:
It should help. After this, I am afraid, `syntax-propertize` is going to be harder to improve.

PR #113, opened 2020-05-22
Recognize async functions when running individual tests

ppletnev:
Tweak regex strings a bit. Without this change, no tests are run with `rustic-cargo-current-test`


brotzeit:
Thanks!

PR #115, opened 2020-05-31
compile-hook

brotzeit:
@tarsius is this something you could do in a hook without issues ?


brotzeit:
Sorry for pinging, but I actually don't use rust for quite some time now and it would be nice to have somebody who takes a look on my changes, so I don't break something again ;)

tarsius:
> I actually don't use rust for quite some time now

I haven't actually really started using rust yet, so I can't really help you test these things.

brotzeit:
Hmmm, ok. Thanks anyway for your help!

PR #116, opened 2020-05-31
respect compilation-scroll-output when first-error

brotzeit:
@tarsius sorry again :D

You can ignore my pings, if you don't want to review my stuff. I know you have lots of other things to do.


tarsius:
> You can ignore my pings, if you don't want to review my stuff.

I plan to look at compilation stuff in another context today. (That's why I finally looked at the breakage first yesterday.) As part of that work I might learn whether this change makes sense.

You should explain in the commit message what issue this fixes and why. And why, if this is the right thing to do, it is not being done by default. *Just from looking at this small diff*, it looks like this is either wrong or it should be done by default without individual each affected packages having to do it.

PR #118, opened 2020-06-03
Add rustic-rustfmt-config-alist.

### LINKED ISSUE #117: Allow rustfmt configuration from Emacs
It would be nice if there were a way to specify arguments to `rustfmt` through a variable.
This would allow the user to make `rustfmt` follow `indent-tabs-mode` or to experiment with formatting options without needing a `.rustfmt.toml`.

Janfel:
This PR adds a `defcustom` called `rustic-rustfmt-args` that specifies a list of arguments to `rustfmt`.
This variable is honored by `rustic-format-buffer` and `rustic-format-start-process`.
I have decided to not make `rustic-cargo-fmt` honor this setting, because this function reformats the entire project and should therefore only obey the project-wide `rustfmt.toml`. Fixes #117.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Rust edition 2018 requires a `rustfmt.toml` file.
 Customization:
 
 - `rustic-rustfmt-bin` path to rustfmt executable
-- `rustic-rustfmt-args` extra arguments to rustfmt
+- `rustic-rustfmt-config-alist` alist of rustfmt configuration options
 - `rustic-format-display-method` default function used for displaying
   rustfmt buffer (use the function `ignore`, if you don't want the
   buffer to be displayed)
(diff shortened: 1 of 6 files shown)


brotzeit:
Can you maybe add a simple test ? I've introduced some bugs recently, so I'm more careful now =)

Janfel:
Sure. I’ve found some inconsistencies in my code as well just now. Is it okay for me to rework the arguments to `rustic-format-start-process` in a (mostly) backwards-compatible way?

brotzeit:
Sure, if you think it makes sense why not.

Janfel:
Just as a heads-up: The commits until now constitute a perfectly serviceable implementation of `rustic-rustfmt-args`, but I have found the experience of actually using such a variable to be lacking.
I have therefore decided to implement the `rustic-rustfmt-config-alist` we were talking about in #117, which provides a much better user experience in my opinion.

brotzeit:
Nice, thanks!

PR #119, opened 2020-06-07
Change Lighter to Rustic

phillord:
Just a suggestion.

Rustic requires rust-mode to be installed, I think. So I have found a little inconsistency in loading (caused by own overly complex set up). Currently, it's not straight-forward to see visually which package is loaded and operational in a buffer.


brotzeit:
Unfortunately, we are still not at the point where rustic uses rust-mode as dependency. But this change makes sense to make.

PR #121, opened 2020-06-14
Remove "-face" suffix from names of faces

tarsius:
As mentioned at (info "(emacs)Defining Faces") "[the name of a face]
should not end in "-face" (that would be redundant).  For historic
reasons a few built-in faces actually do end with "-face" and it so
happens that many of our faces are closely related to just those
`font-lock` and `compilation` faces and that probably inspired our
use of the suffix.  Even so, we should stop now.

PS: I did not base this on the latest tip of `master` because there is a recent regression, which I did not wish to investigate.  `make` should tell you about that issue.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -31,7 +31,10 @@ If nil then the project is simply created."
   "Face used for outdated crates."
   :group 'rustic)
 
-(defface rustic-cargo-outdated-upgrade-face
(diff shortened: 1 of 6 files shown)


tarsius:
I'll look into the test failure.

tarsius:
I have updated the tests and now they pass again.

brotzeit:
Thanks!

PR #125, opened 2020-07-01
fixed: restore char point when rustfmt error

ncaq:
`rustic-format-file-sentinel` function restore char point when Rustfmt exit on success.
So when first save and rustfmt cause error, char-point move to buffer start.
I fix to restore point process and when rustfmt cause error.


brotzeit:
Thanks!

PR #126, opened 2020-07-03
allow sending input to process with rustic-compile-send-input

brotzeit:


l2dy:
I got a `rustic-compile-send-input: Wrong number of arguments: #<subr process-send-string>` error, is it a bug?

brotzeit:
I just forgot an argument.

Issue #143, opened 2020-09-17
helm-ag should be optional for rustdoc


brotzeit:
We need a defcustom to define which command is to be used to run ripgrep.

brotzeit:
It would also be good to mention that lsp-mode is required for this feature. Even better would be if it also would work with eglot.

brotzeit:
It would be ideal if it would also be usable with a command that's provided by emacs. Generally it's always good to reduce hard dependencies.

brotzeit:
@samhedin this issue can be closed right ?

samhedin:
Yes, this can be closed. I will continue to work on making the installation easier, but the time I have is spontaneous at the moment since school started again. Looking forward to it though :)

PR #144, opened 2020-09-17
Defvar search command, make helm-ag dependency even more optional.

samhedin:
For https://github.com/brotzeit/rustic/issues/143
This extracts the search functionality to the variable `rustdoc-search-function` so that the user can change it. I was unsure whether or not variables which are functions are reasonable subjects for defcustom, but if they are I will change it.
Furthermore, if helm-ag and ripgrep are not installed the search function is set to the emacs grep command, using grep.


brotzeit:
You can use a function instead and add a defcustom similiar to:

```elisp
(defcustom rustic-rustdoc-search-method 'with-helm-ag
  "Default function used for ..."
  :type 'function
  :group 'rustic-rustdoc)
```

and then actually call the function with `funcall`.

rustic-rustdoc.el
```diff
@@ -50,6 +49,20 @@ All projects and std by default, otherwise last open project and std.")
                              ()
                              ,(concat rustdoc-source-repo "filter.lua"))))
 
+(defvar rustdoc-search-function (if (and  (require 'helm-ag nil t) (executable-find "rg")) ; If helm-ag and rg is available we use them by default, otherwise revert to using grep
+                                       (lambda (search-dir search-term)
+                                         (let* ((helm-ag-base-command (if rustdoc-current-project ; If the user has not visited a project the search will be done from the doc root, in which case we should not follow symlinks.
+                                                                          "rg -L --smart-case --no-heading --color=never --line-number --pcre2"
+                                                                        "rg --smart-case --no-heading --color=never --line-number --pcre2"))
```
brotzeit, line 56: The command should also be a `defcustom`. And maybe we can avoid the duplication and use `concat` instead.

rustic-rustdoc.el
```diff
@@ -50,6 +49,20 @@ All projects and std by default, otherwise last open project and std.")
...
+                                                (helm-ag-fuzzy-match t)
+                                                (helm-ag-success-exit-status '(0 2)))
```
brotzeit, line 58: Is this really necessary for this functionality ? These options should generally be chosen by the users.

rustic-rustdoc.el
samhedin: Done and done. Thank you.

rustic-rustdoc.el
samhedin: The first one, no. I believe that is a leftover from me playing around with the settings. The second one is from the documentation of helm-ag: "Ripgrep uses exit-status 2 to indicate a partial success:

(custom-set-variables
 '(helm-ag-base-command "rg --no-heading")
 `(helm-ag-success-exit-status '(0 2)))"

I did not set the exit status at first but ran into some surprising behavior which I can't remember now, and subsequently added it.

samhedin:
> You can use functions instead and add a defcustom similiar to:
> 
> ```emacs-lisp
> (defcustom rustic-rustdoc-search-method 'with-helm-ag
>   "Default function used for ..."
>   :type 'function
>   :group 'rustic-rustdoc)
> ```

Done, with slightly different naming. Would you prefer rustic-rustdoc over rustdoc? I see the point of the former, but it is quite long.

brotzeit:
> Would you prefer rustic-rustdoc over rustdoc?

It's a convention.

brotzeit:
Switching to `doc` definitely makes sense :+1:

brotzeit (approved):

samhedin:
Yeah, this feels more natural! Just need to smooth out some issues with regular grep now and then it should be ready.

brotzeit:
:+1:

PR #147, opened 2020-09-25
add cargo-edit commands

### LINKED ISSUE #145: Feature Request: Integration with Cargo Edit
This exists: https://lib.rs/crates/cargo-edit

which manages the .toml file nicely. `Rust-mode` comes packaged with `cargo-edit` commands. Easy ask would be to integrate `rustic mode` with pre-packaged keybinds. 
Slightly more involved ask would be to have `rustic-mode` run cargo add on any packages mentioned in the `use blah` list.

brotzeit:
close #145


samhedin (commented):
Looks good, though I think there could be a point to informing the user that they need to install `cargo edit` for these commands to work, outside of the README. 
A user might do `M-x rustic-` to see what commands are available, think that `rustic-cargo-add` looks super, run it and be greeted with an error complaining about `cargo add` missing. This is doubly confusing since `cargo add` actually looks like a command that might be included in `cargo` by default, so the initial thought might be that `rustic` is broken.

A notice about the user needing to install `cargo edit` in the documentation strings would be helpful. Or better yet, prompt them to install it if they run the command without `cargo edit` installed.

brotzeit:
Done.

samhedin (approved):

PR #159, opened 2020-09-30
use :stderr for formatting buffers

brotzeit:


samhedin:
When I run `rustic-format-buffer` on an incorrectly formatted buffer, an empty buffer called `*rustfmt* shows up. I assume that the buffer that should show up is `*rustfmt-warnings*`? It also appears but in the background.

brotzeit:
@samhedin should work now.

samhedin:
Using it prints `rustic-format-buffer: No buffer named *rustfmt-warnings*`.

brotzeit:
@samhedin can you please test it again. I don't see the problem...

samhedin:
Running `rustic-format-buffer` on a buffer without syntax errors works great now! 
But when I try with this 
```rust
#![allow(unused_imports, unused_variables, dead_code)]
fn main() {
}}
}
```
The empty buffer `*rustfmt*` shows up. In the background, the filled buffer `*rustfmt-warnings*` also appears with the actual warning. 
```
error: unexpected closing delimiter: `}`
 --> <stdin>:3:2
  |
2 | fn main() {
  |           - this opening brace...
3 | }}
  | -^ unexpected closing delimiter
  | |
  | ...matches this closing brace


Process rustic-rustfmt-process stderr finished
```
Are you seeing different behavior? I should have been more clear about this, no idea why I didn't specify what issues I was having when formatting a correct vs incorrect buffer.

brotzeit:
I just tried it with warnings and not with an error.

But I know what the issue is. I'll try to fix it.

brotzeit:
Now I remember why I didn't use `err-buf` in the first place. It seems it's not possible to add a filter function for the process stderr output. I really like the colored output...

samhedin:
Does this mean that you would like to keep the current behavior and that this PR should be closed?

brotzeit:
I wonder how hard it would be to add this functionality and get it merged.

Issue #150, opened 2020-09-25
cargo-add: add feature to automatically detect missing dependencies


brotzeit:
The idea is to parse crates from use statements(in the current buffer) and look them up in Cargo.toml. Then the user could get a minibuffer prompt containing the missing dependencies similar to `Add missing dependencies: cargo add crate1 crate2 crate3`.

#145

samhedin:
Would it be easier to rely on the `unresolved import` error rather than parsing manually?
This would require the user to have a language server, of course.

brotzeit:
Sounds reasonable. Most people use lsp anyway.

samhedin:
Closing this as #161 was merged.

brotzeit:
Unfortunately we can't merge this until it also supports eglot or it will only be loaded when lsp-mode is present.

samhedin:
You're right, I was too quick on the trigger. I'll take a look at eglot support!

brotzeit:
Should be working now.

PR #161, opened 2020-10-01
Add missing dependencies

samhedin:
For https://github.com/brotzeit/rustic/issues/150
`missing-imports` is just a list at the moment. What is the preferred way to prompt the user to select one of them? An alternative would be to just add all missing dependencies. I'm not sure how often a user would input several missing dependencies and then only want to add a subset of them. And in those cases, you could just not run `rustic-cargo-add-missing-import`.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -446,7 +446,7 @@ If running with prefix command `C-u', read whole command from minibuffer."
                       (concat "cargo add " (read-from-minibuffer "Crate: ")))))
       (rustic-run-cargo-command command))))
 
-(defun rustic-add-missing-import ()
+(defun rustic-cargo-add-missing-import ()
   "Add missing imports to Cargo.toml."
   (interactive)
   (let ((missing-imports (seq-reduce (lambda (missing-crates errortable)
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -446,6 +446,13 @@ If running with prefix command `C-u', read whole command from minibuffer."
                       (concat "cargo add " (read-from-minibuffer "Crate: ")))))
       (rustic-run-cargo-command command))))
 
+(defun rustic-add-missing-import()
+    "Add missing imports to cargo.toml."
+  (interactive)
+  (dolist (errortable (gethash (buffer-file-name) (lsp-diagnostics)))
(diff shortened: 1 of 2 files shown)


brotzeit:
I would say we should add all missing dependencies by default. When the user runs the command with prefix argument it should offer the user dependencies that he can choose manually.

samhedin:
Good idea, I added it now.

rustic-cargo.el
```diff
@@ -446,6 +446,22 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
+Adds all missing imports by default.
+Use with 'prefix-arg` to select imports to add."
+  (interactive)
+  (let ((missing-imports (delete-dups
+                          (seq-reduce (lambda (missing-crates errortable)
+                                        (if (string= "E0432" (gethash "code" errortable))
+                                            (cons (nth 3 (split-string (gethash "message" errortable) "`")) missing-crates)
+                                          missing-crates))
+                                      (gethash (buffer-file-name) (lsp-diagnostics t)) '()))))
+    (rustic-run-cargo-command (format  "cargo add %s"
```
brotzeit, line 460: Why not use `rustic-cargo-add` ?

samhedin: `rustic-cargo-add` reads from the minibuffer, so I figured it was not meant for the case that's needed here. Is there a way to use `rustic-cargo-add` that I am missing? Of course, `rustic-cargo-add` could be modified to work with `rustic-cargo-add-missing-import`, but I think the end result would be worse than the current situation.

brotzeit: You're right, I haven't thought about it.

I think we should also handle the case when `missing-imports` is `nil`.

samhedin: Done!

brotzeit:
Very handy. If you want you could also use `if-let`.

brotzeit (approved):
You can merge when you're happy with it.

brotzeit:
Ah sorry, please add a note to the readme. And please mention that this function currently only works with lsp-mode.

samhedin:
Good catch!

samhedin:
> Very handy. If you want you could also use `if-let`.

Oh, I missed this comment. You are right, this would have been nicer. I am looking into eglot support, will fix there.

PR #175, opened 2020-10-27
Use Tramp aware calls

c-nixon:
This tweaks a couple of calls so that rustic works with cargo, etc on a remote machine.  As far as I can tell it doesn't have any effect on a more conventional local setup.


brotzeit:
I think I have also considered using `start-file-process`, but I don't remember why I chose `make-process`. Give me some time to think about it maybe I can remember my intentions ;)

brotzeit:
I think this is the best solution. Thanks for the PR!

PR #177, opened 2020-10-30
Missing imports eglot

samhedin:
For https://github.com/brotzeit/rustic/issues/169
Now supports both eglot and lsp. I'm not a huge fan of how I collect the missing imports from eglot. It works by inspecting errors that eglot is about to send to flymake, which makes it a bit hacky. In the lsp-mode version, a call to the function `lsp-diagnostics` just returns everything that's needed, but I could not find an equivalent function in eglot, though obviously it must be there somewhere. 

Werks on my machine, but I need to try it a bit more and have another look at the code.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -472,7 +472,7 @@ Adds all missing imports by default.
 Use with 'prefix-arg` to select imports to add."
   (interactive)
   (when (rustic-cargo-edit-installed-p)
-    (if-let ((missing-imports (if lsp-mode
+    (if-let ((missing-imports (if (ignore-error lsp-mode)
                                   (rustic-cargo--lsp-missing-imports)
                                 (rustic-cargo--eglot-missing-imports))))
         (rustic-run-cargo-command (format  "cargo add %s"
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -447,6 +447,7 @@ If running with prefix command `C-u', read whole command from minibuffer."
       (rustic-run-cargo-command command))))
 
 (defun rustic-cargo--eglot-missing-imports ()
+  "Find missing imports when using eglot."
   (delete-dups (seq-reduce (lambda (missing-crates clstruct)
                              (if (and  (flymake--diag-p clstruct)
                                        (string-match-p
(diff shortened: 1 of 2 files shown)


brotzeit:
We still have the problem that we call lsp-mode/flymake code that we didn't require.

samhedin:
Good catch, in my mind `lsp-mode` would return nil when undefined for some reason, but this is clearly not how most languages work. The flymake one is worse though. I want to remove the dependency on it altogether, but the eglot source code is doing so many things I'm unfamiliar with that I've been unable to get it to work with it at all. I might contact the author if I can't come up with anything.

brotzeit:
Maybe we can add files for eglot and lsp-mode and require these files somewhere.

samhedin:
As it is now, LSP and eglot should not be required if they're not loaded thanks to `(ignore-error lsp-mode)`. Is there another reason to split them up?

rustic-cargo.el
```diff
@@ -446,6 +446,42 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
+                   missing-crates))
+               (gethash (buffer-file-name) (lsp-diagnostics t)) '())))
+
+(defun rustic-cargo-add-missing-imports ()
+  "Add missing imports to Cargo.toml.
+Adds all missing imports by default.
+Use with 'prefix-arg` to select imports to add."
+  (interactive)
+  (when (rustic-cargo-edit-installed-p)
+    (if-let ((missing-imports (if (ignore-error lsp-mode)
```
brotzeit, line 475: You can do something like `(if (featurep 'lsp-mode) (print "foo"))`

rustic-cargo.el
```diff
@@ -446,6 +446,42 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
+(defun rustic-cargo--eglot-missing-imports ()
+  "Find missing imports when using eglot."
+  (delete-dups (seq-reduce (lambda (missing-crates clstruct)
+                             (if (and  (flymake--diag-p clstruct)
+                                       (string-match-p
+                                        (regexp-quote "unresolved import `")
+                                        (format "%s" (flymake--diag-text clstruct))))
+                                 (cons (nth 3 (split-string (flymake--diag-text clstruct) "`")) missing-crates)
+                               missing-crates))
+                           eglot--unreported-diagnostics '())))
```
brotzeit, line 458: Here we also have to funcall.

samhedin:
Just an update, I contacted the eglot maintainers and they'd rather I do something else than the current solution, so I will look into that when I get time, but school is taking up all my focus at the moment.

Issue #174, opened 2020-10-26
rustic-buffer-workspace does not find the workspace root


mullr:
If you have workspace with multiple crates, it gives you the directory of the crate which contains buffer's file, not the workspace root. The name and docstring indicate that this may not be intended.

samhedin:
It might be best to rely on the field `workspace_root` with the built in `cargo metadata` command.

brotzeit:
I hope this helps #294

PR #179, opened 2020-10-31
Use cargo metadata to find workspace root

samhedin:
https://github.com/brotzeit/rustic/issues/174
How about this?

### DIFF
--- a/rustic-common.el
+++ b/rustic-common.el
@@ -59,9 +59,12 @@ non-nil."
   "Get the workspace root.
 If NODEFAULT is t, return nil instead of `default-directory' if directory is
 not in a rust project."
-  (let ((dir (alist-get '\workspace_root
-                        (json-read-from-string
-                         (shell-command-to-string "cargo metadata --format-version 1")))))
+  (let ((dir (or  (condition-case nil
+                      (alist-get '\workspace_root
+                                 (json-read-from-string
+                                  (shell-command-to-string "cargo metadata --format-version 1")))
+                    (error nil))
+                  default-directory)))
     (if dir
         (expand-file-name dir)
       (if nodefault
--- a/rustic-common.el
+++ b/rustic-common.el
@@ -60,7 +60,7 @@ non-nil."
 If NODEFAULT is t, return nil instead of `default-directory' if directory is
 not in a rust project."
   (let ((dir (locate-dominating-file
-              (or buffer-file-name default-directory) "Cargo.toml")))
+              (or buffer-file-name default-directory) "Cargo.lock")))
     (if dir
         (expand-file-name dir)
       (if nodefault
--- a/rustic-common.el
+++ b/rustic-common.el
@@ -54,7 +54,6 @@ non-nil."
   (or rustic-format-on-save (eq rustic-format-trigger 'on-save)))
 
 ;;; Essentials
-
 (defun rustic-buffer-workspace (&optional nodefault)
   "Get the workspace root.
 If NODEFAULT is t, return nil instead of `default-directory' if directory is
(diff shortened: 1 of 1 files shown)


samhedin:
@brotzeit it appears that these builds are failing because some tests use `string=` to compare paths. Would it be appropriate to change these tests to use https://github.com/rejeep/f.el#f-same-path-a-path-b

brotzeit:
I think we once used this kind of code in rust-mode and somebody changed it to the current version. I hesitated to use it since I wanted to avoid the use of a shell process. I still would like to avoid it if possible.

samhedin:
That's fair. From what I gather from the documentation, simply searching for `Cargo.lock` instead of `Cargo.toml` should take care of it.

brotzeit:
@mullr can you give it a try ?

samhedin:
I should add that I did try it on the basic case listed here: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
This is the first time I touched workspaces so I'm not certain my test was exhaustive.

mullr:
@brotzeit This seems to be working for me now. Thanks for fixing it!

mullr:
It just occurred to me that this new method may not be entirely correct either. I think there are two axes of configuration here, for idiomatic Rust projects:

1) Standalone crate
2) Multiple crates in a workspace

A) Cargo.lock is checked in (because this is producing a binary)
B) Cargo.lock is not checked in (because this is a producing a library)

The new method gives incorrect results for 2B when you have a fresh checkout. This is difficult to work around; one method that comes to mind would be to keep looking up the tree for Cargo.lock (I *think* this is what cargo does itself, but it's probably worth checking). 

It's also worth noting that the old method happened to be more correct (but misnamed) for specific test case execution, which now will attempt to run the named test at the workspace level. This is often just a little slower (since it runs every test), but could run a different test if you have one with the same name in a different crate in the workspace.

samhedin:
> one method that comes to mind would be to keep looking up the tree for Cargo.lock
Is this not covered by the use of `locate-dominating-file`? If not, we can, as you say, have a closer look at how cargo does it and try to mimic it.

Thanks!

mullr:
> Is this not covered by the use of locate-dominating-file? If not, we can, as you say, have a closer look at how cargo does it and try to mimic it.

Unfortunately it isn't; `locate-dominating-file` stops at the first instance of the given file it finds. 

I did some experimentation and a bit of research, and I'm afraid the it's more complicated than I thought: it doesn't look like it's possible to do this directly without reading the file content. Notably, it's perfectly valid to have one crate directory nested inside another, but /not/ coexisting in the same workspace. This is indistinguishable from a workspace until you examine the file contents. 

I put together the test scenarios I could think of here: https://github.com/mullr/rustic-workspace-test-scenarios

I think this is the code cargo uses: https://github.com/rust-lang/cargo/blob/2af662e22177a839763ac8fb70d245a680b15214/src/cargo/core/workspace.rs#L430

samhedin:
Alright, maybe mimicking that is the best way. I can't make a new PR though, I'm super busy with non-emacs things in life at the moment.

mplanchard:
Hi and thanks for an excellent project!

I just wanted to note for anyone else who winds up here from a search engine that I was running into an issue where  on a large workspace, the failure to resolve the workspace root correctly was causing new `rust-analyzer` instances to spin up for each sub-project. This eventually caused problems, because each rust-analyzer instance would use a pretty significant amount of RAM, and eventually it would cause system swapping and general sadness.

I was able to resolve the issue by adding a `.dir-local.el` file to the root of the project containing:

```elisp
((nil . ()) 
 (rustic-mode ;; for rustic mode
  . ((eval . (setq-default rustic--buffer-workspace (projectile-project-root))))))
```

Probably not ideal, but it's fixed things for me here.

Issue #176, opened 2020-10-27
How to use semver in org-babel crate definition?


0atman:
I can't seem use the `time` crate's current version in an Babel block:

```rust
#+begin_src rust :crates '((time . 0.2.22))
// cargo-deps: time
extern crate time;
fn main() {
    println!("{}", time::now().rfc822z());
}
#+end_src
```

`Wrong type argument: symbolp, "time"`

0atman:
I can confirm that the example works for me:

```rust
#+BEGIN_SRC rust :crates '((regex . 0.2))
extern crate regex;

use regex::Regex;

fn main() {
    let re = Regex::new(r"^\d{4}-\d{2}-\d{2}$").unwrap();
    println!("{:?}", re.is_match("2014-01-01"));
}
#+END_SRC

#+RESULTS:
: true
```

0atman:
The difference between the two seems, to me, to be the float version number for `regex`, and the semver for `time`?

0atman:
I THINK that this can never work with semver crates, because it runs `number-to-string`, which expects an int or float: https://github.com/brotzeit/rustic/blob/master/rustic-babel.el#L220

brotzeit:
Yeah, I supposed this was the case. I guess we have change the version number to be a string.

0atman:
@brotzeit could it be as simple as https://github.com/brotzeit/rustic/pull/180 ? :grin: 

Full disclosure, this is the first elisp I've written!

PR #180, opened 2020-11-02
crate version should be a string

0atman:
removed number-to-string

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -217,7 +217,7 @@ directory DIR."
         (dependencies ""))
     (dolist (crate crates)
       (let ((name (symbol-name (car crate)))
-            (version (number-to-string (cdr crate))))
+            (version (cdr crate)))
         (setq dependencies (concat dependencies name " = " "\"" version "\"" "\n"))))
     (setq dependencies (concat "[dependencies]\n" dependencies) )
     (make-directory (file-name-directory toml) t)


samhedin:
Apologies if I'm a bit out of the loop, but I'm giving it a try anyway.
When running the first snippet posted in https://github.com/brotzeit/rustic/issues/176 I get `rustic-babel-cargo-toml: Wrong type argument: numberp, 0\.2\.22`
``` rust
#+begin_src rust :crates '((time . 0.2.22))
// cargo-deps: time
extern crate time;
fn main() {
    println!("{}", time::now().rfc822z());
}
#+end_src
```
Is this working for you?

0atman:
I fully don't expect my code to work, as I'm not an elisp dev, nor do I know how to run it.
But the semver should be a string in your example. That's what I hoped my fix would support.
`0.2.22` isn't valid syntax, numerically speaking, right? :grin:

samhedin:
Of course, got it. `rustic-test-babel-crate` fails now, but it should not be too bad to fix. If you can update the test accordingly, I say this looks good to merge!

brotzeit:
I think it's better to keep things backwards compatible. Can you guys please take a look at #181

brotzeit:
@0atman thanks for your help. Next time we will guide you to the solution and merge your PR.

PR #181, opened 2020-11-06
allow strings for crate versions

brotzeit:


samhedin:
I tried it and it works. It would have been fun to merge the original PR, but this looks good.

0atman (approved):
Smart to keep the old behavior as an option 👌

brotzeit:
> It would have been fun to merge the original PR, but this looks good.

Yeah, sorry. I'm currently focused on https://github.com/remacs/emacs-webrender and didn't think about it.

PR #183, opened 2020-11-11
Silence byte-compiler warnings

leungbk:
----

#

### DIFF
--- a/rustic-doc.el
+++ b/rustic-doc.el
@@ -60,6 +60,10 @@ Needs to be a function because of its reliance on
   :type 'function
   :group 'rustic-doc)
 
+(defvar helm-ag-base-command)
+(defvar helm-ag-success-exit-status)
+(declare-function helm-ag "ext:helm-ag")
+
 (defun rustic-doc-default-search-function (search-dir search-term)
   "Default search functionality.
 Uses helm-ag and ripgrep if possible, grep otherwise.


samhedin:
``` emacs-lisp
(defvar helm-ag-base-command)
(defvar helm-ag-success-exit-status)
```
This will make the variables dynamically scoped. The result will be that they will still be set to `rustic-doc-` specific values after rustic-doc is finished, meaning that if you're using `helm-ag` for something else you will see different behavior if you've used rustic-doc. Some initial tests seem to confirm this. Is there a different way to silence the warnings?

leungbk:
> From my understanding this will make the variables dynamically scoped. The result will be that they will still be set to `rustic-doc-` specific values after rustic-doc is finished, meaning that if you're using `helm-ag` for something else you will see different behavior if you've used rustic-doc. 

Outside the [`let` expression](https://github.com/brotzeit/rustic/blob/af437a2ab2b92a8a946b25e3ecf7a970b8316230/rustic-doc.el#L73), they should revert to the values that they had before. I tried running `rustic-doc-search` within a project, then after completing the search, ran `M-x describe-variable` and found that the variables went back to their old values.

What do your tests look like?

samhedin:
You are correct - I am seeing some strange behavior all around on emacs when using my desktop, which I recently switched distro on. I'm not sure how that impacts this, but when I gave it a try on my laptop all worked beautifully. 
Thanks for the PR, I didn't actually believe that anyone was using `rustic-doc` but now there's been a PR and an issue in the last week which makes me very happy.

samhedin (approved):

PR #184, opened 2020-11-13
Add async as a dependency

leungbk:
`async` is used by `rustic-doc.el`.

Also silence a separate byte-compiler warning.


brotzeit:
How could I miss this...Thanks!

PR #188, opened 2020-11-20
No longer depend on async and projectile

tarsius:
Don't depend on `async` because doing so is completely unnecessary and don't depend on `projectile` because it should not be forced on users for the sake of some minor functionality (which we can also have without using `project` instead).

### DIFF
--- a/rustic-compile.el
+++ b/rustic-compile.el
@@ -10,9 +10,9 @@
 ;;; Code:
 
 (require 'xterm-color)
-(require 'projectile)
 (require 'markdown-mode)
 
+(require 'cl-lib)
 (require 'dash)
 (require 'subr-x)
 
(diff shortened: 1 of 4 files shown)


brotzeit:
Thanks, I'll try to fix the text.

brotzeit:
It seems the buffers aren't reverted.

tarsius:
Sorry about that. I've pushed a fixed version.

Sadly `project--buffer-list` has to be wrapped.

tarsius:
There was one more problem as I've been told by the tests.
I fixed it and now the tests are... gone?
Anyways. Please have another look.

brotzeit:
Yeah travis is the problem. I will run the tests manually for now.

tarsius:
> Yeah travis is the problem.

I'll probably have to deal with that soon too. :frowning_face:

brotzeit:
:/

brotzeit:
Oh nice green :+1:

tarsius:
Sadly that required backporting. It's weird these things didn't exist until now.

PR #195, opened 2020-12-02
rustic-rustfix: change function call

leungbk:
----

#


brotzeit:
Thanks!

The build isn't actually failing, it seems we have to move away from travis

 > Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.

PR #200, opened 2020-12-12
rustic-cargo-run: minibuffer and history

matzebond:
This reads arguments for the `cargo run` command from the minibuffer and
saves the history.

I was missing this feature when switching form the default `compile` or
`projectile-run-project`.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -387,9 +387,6 @@ If BIN is not nil, create a binary application, otherwise a library."
   (interactive)
   (rustic-run-cargo-command "cargo build"))
 
-(defvar rustic--cargo-run-argument-history nil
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks! Do you think it's important that `cargo run` gets its own history or would it be ok to use `compile-history` ?

matzebond:
> Thanks! Do you think it's important that `cargo run` gets its own history or would it be ok to use `compile-history` ?

You are right it should be `compile-history`, I just wasn't aware of it.

brotzeit:
Can you please use `rustic-cargo-bin` instead of 'cargo'. I have to finish the PR that takes care of the remaining commands.

brotzeit:
Ah sorry didn't notice you already updated.

PR #203, opened 2020-12-18
Fix rustic-cargo-new

monoid:
1. Make it ask any path, not an existing directory.
2. Apply path expansion, otherwise "~/blabla" is passed directly to
   cargo new, and it silently creates "~" directory in the current one.


brotzeit:
I wonder why this actually works(without your changes). The sentinel shouldn't be able to read `project-path`, at least for me that's the case.

monoid:
Well, it doesn't :)
rustic-new

PR #214, opened 2021-01-09
Fix rustic-cargo-fn-regexp

### LINKED ISSUE #213: Slight glitch in rustic-cargo-fn-regexp
Consider in the following code `(rustic-carg--get-current-fn-name)` evaluates to `nil` when `(point)` is inside `fn foobar()`

``` rust
mod foo {
    fn bar() {
    // (rustic-cargo--get-current-fn-name) evals to bar ✓
    }
}

fn foobar () {
// (rustic-cargo--get-current-fn-name) evals to nil ❌
}
```

johannes-mueller:
Fixes #213


johannes-mueller:
Will be superseeded by PR with tests adjusted

PR #219, opened 2021-01-29
Adding crate feature support to org-babel :crates arg

tylerware:
I wanted to use a feature of a crate today and could not with the `:crates` header arg for org-babel. Anyway, this adds support for the following to work 

```orgmode
#+begin_src rust :crates '((rand . "0.7") (num-bigint . ((version . 0.3) (features . rand))))
use num_bigint::RandBigInt;
fn main() {
    let mut rng = rand::thread_rng();
    println!("Big random number: {}", rng.gen_biguint(2048 as u64));
}
#+end_src
```

It also supports this format to `(features . (rand))` for multiple features for a crate. I ran the tests with this result:
```
Ran 245 tests in 32.071 seconds
1 expected failures
```

I didn't spend a whole lot of time on it, so if you want changes just let me know!

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -218,10 +218,8 @@ directory DIR."
         (dependencies ""))
     (dolist (crate crates)
       (let ((name (symbol-name (car crate)))
-            (version (cdr crate)))
-        (when (numberp version)
-          (setq version (number-to-string version)))
-        (setq dependencies (concat dependencies name " = " "\"" version "\"" "\n"))))
+            (version (rustic-babel-generate-version-toml (cdr crate))))
+        (setq dependencies (concat dependencies name " = " version "\n"))))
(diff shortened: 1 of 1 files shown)


brotzeit:
Emacs master doesn't signal an error for some tests with `should-error`. This looks familiar, maybe we should just turn the tests off hmmm.

brotzeit:
Ok it seems it was just another temporary failure(happens frequently). But it seems we have another possible solution with #220. I have to take a look...

brotzeit:
Thanks for your pull request, but I'll going to merge the other one. If there are tests present it's a lot easier for me to review changes. It frequently happens that I merge pull requests which introduce bugs so this is a huge advantage for me. Additional pull requests are always very appreciated.

PR #222, opened 2021-02-07
Fix indentation of chained methods

j0ni:
Method calls after `?` operators should be indented too.

This is a trivial fix, and I didn't go looking for a better way than just adding a second call to `rustic-looking-back-str`. I'll be just as happy if you just treat this as a bug report :)


j0ni:
I don't think this does what I thought it did, and I'm not sure how I came to believe it did. Anyway, closing this and will likely make a bug report instead.

brotzeit:
Ok.

Issue #227, opened 2021-02-19
rustic-doc-setup fails silently on macOS: nproc not present


antifuchs:
I am just trying out rustic-mode, and would love to use its convenient doc lookup; unfortunately on macOS (without a `coreutils` installation), there is no `nproc` program, which makes `~/.local/bin/rustic-doc-convert.sh` fail:

```
:;    /Users/asf/.local/bin/rustic-doc-convert.sh  std
Generating org files in: /Users/asf/.local/share/emacs/rustic-doc/std
/Users/asf/.local/bin/rustic-doc-convert.sh: line 51: nproc: command not found
error: The argument '--threads <num>' requires a value but none was supplied

USAGE:
    fd [FLAGS/OPTIONS] [<pattern>] [<path>...]

For more information try --help
:;    echo $?
1
```

What's even more concerning though, is that running `M-x rustic-doc-setup RET` succeeds. Here's an excerpt from my `*Messages*` buffer:

```
Contacting host: raw.githubusercontent.com:443
Wrote /Users/asf/.local/bin/rustic-doc-convert.sh
Wrote /Users/asf/.local/bin/rustic-doc-filter.lua
Setup is converting the standard library
Finished converting docs for std
```

I'm going to install coreutils to get the `nproc` program (would be good to have anyway), but it would be really useful if the setup function could error if that conversion tool exits non-0, instead of pretending that everything is alright.

samhedin:
> it would be really useful if the setup function could error if that conversion tool exits non-0, instead of pretending that everything is alright.
Completely agreed. In addition, maybe we could add a more macos native function call there instead of relying on the user installing nproc. `sysctl -n hw.logicalcpu` is apparently one way to do it, but I can't try it out as I don't have access to a mac.

antifuchs:
`sysctl -n hw.logicalcpu` on my 8-core macbook returns 16, same as `nproc` from coreutils does - that would be a viable solution!

samhedin:
@antifuchs could you give it a try and confirm that everything works on macos? If so we can close this

samhedin:
Closing for now :/)

PR #228, opened 2021-02-20
Report error and better way to find number of cores on macos 

samhedin:
For https://github.com/brotzeit/rustic/issues/227
The error reporting is a bit... primitive and non-informative. Is there really no better way to check if the event to process sentinel indicates error than string comparison?

I want to look more into this, and also see if it can work on windows, but for now, it can be tried on macos to see if this works.


rustic-doc.el
```diff
@@ -287,14 +288,17 @@ If the user has not visited a project, returns the main doc directory."
...
     (set-process-sentinel
-     proc (lambda (proc _event)
+     proc (lambda (proc event)
             (let ((buf (process-buffer proc)))
-              (when finish-func
-                (funcall finish-func proc))
-              (when (buffer-live-p buf)
-                (kill-buffer buf)))))
+              (if (string-match-p (regexp-quote "abnormally") event)
+                  (message "Could not finish process: %s. See *Messages* or a *rustic-*...* buffer for more info." event)
```
brotzeit, line 294: > or a *rustic-*...*

You mean the process buffer ?

samhedin: Right, I wasn't sure what it should be called but it seems like it's not worse than what's there now

samhedin:
I added a gif of rustic-doc in use. Keep it, replace with image or toss it away?

rustic-doc.el
```diff
@@ -287,14 +288,17 @@ If the user has not visited a project, returns the main doc directory."
...
     (set-process-sentinel
-     proc (lambda (proc _event)
+     proc (lambda (proc event)
             (let ((buf (process-buffer proc)))
-              (when finish-func
-                (funcall finish-func proc))
-              (when (buffer-live-p buf)
-                (kill-buffer buf)))))
+              (if (string-match-p (regexp-quote "abnormally") event)
+                  (message "Could not finish process: %s. See the *Messages* buffer or %s for more info." event (concat "*" name "*"))
```
brotzeit, line 294: You can't access `name` from the sentinel.

samhedin: I try like this: 
* Add `exit 1` to the top of `/home/sam/.local/bin/rustic-doc-convert.sh`.
* Comment out the line ` (rustic-doc--install-resources)` in `rustic-doc-setup` inside `rustic-doc.el` (otherwise the script would * run `rustic-doc-setup`
This gives the message
`Could not finish process: exited abnormally with code 1
. See the *Messages* buffer or *rustic-doc-std-conversion* for more info.`

So it seems to me that I can access the name in the sentinel, or am I misunderstanding something?

brotzeit: Ah you're right. You can capture the variable if you use a lambda. If you would use a function instead of the lambda it wouldn't work. I'm not coding much in elisp for some time.

brotzeit (approved):

rustic-doc.el
samhedin: It's good to make sure. Thank you!

PR #232, opened 2021-03-14
Use generic function project-roots to retrieve the roots instead of calling cdr directly

hectorhon:
Use generic function project-roots to retrieve the roots instead of calling cdr directly

Some setups add a custom function to project-find-functions, so the generic function project-roots should be used instead of directly calling cdr on the result of project-current.

### DIFF
--- a/rustic-util.el
+++ b/rustic-util.el
@@ -247,7 +247,9 @@ This is basically a wrapper around `project--buffer-list'."
     (if (fboundp 'project--buffer-list)
         (project--buffer-list pr)
       ;; Like the above function but releases before Emacs 28.
-      (let ((root (car (project-roots pr)))
+      (let ((root (if (>= 28 emacs-major-version)
+                      (project-root pr)
+                    (car (project-roots pr))))
             bufs)
         (dolist (buf (buffer-list))
           (let ((filename (or (buffer-file-name buf)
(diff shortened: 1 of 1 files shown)


hectorhon:
Yeah, that should work too, feel free to edit. I'm on vanilla Emacs 27.1 and don't have the function project-root defined... It must have been added in a new version of project.el.

samhedin:
I did not realize this. Apparently this was changed as recently as last year.
This should work, but I haven't tested it (can do later).
``` emacs-lisp
(root (if (>= 28 emacs-major-version)
                      (project-root pr)
                    (car (project-roots pr))))
```

Interestingly, the definition of `project-root` is the same as you did it:
``` emacs-lisp
(cl-defgeneric project-root (project)
  "Return root directory of the current project.

It usually contains the main build file, dependencies
configuration file, etc. Though neither is mandatory.

The directory name must be absolute."
  (car (project-roots project)))
```
But it might still be worthwhile to switch to using `project-root` for the sake of not relying on obsolete functionality.

samhedin (approved):

samhedin:
Okay, so it wasn't as straightforward as I had hoped. I think this is good for now, so I'm merging. Thank you!

hectorhon:
Thank you @samhedin !

PR #239, opened 2021-03-31
Use cl-lib exclusively

tarsius:
The first commit fixes a regression; something started using `cl-lib`
without making sure it was loaded.  That broke "make lisp" and by
extension "make test".

Once testing worked again I noticed that one test file still required
`cl` instead `cl-lib`.  Once I had fixed that using the second commit,
one test started to fail:

``` emacs-lisp
Test rustic-test-backtrace condition:

    (ert-test-failed
     ((should
       (= compilation-num-errors-found 1))
      :form
      (= 0 1)
      :value nil))

```

I have absolutely no idea why that is. Someone else will have to
figure it out.

### DIFF
--- a/test/rustic-rust-mode-test.el
+++ b/test/rustic-rust-mode-test.el
@@ -1,20 +1,20 @@
 ;; -*- lexical-binding: t -*-
 
-(require 'cl)
+(require 'cl-lib)
 (require 'imenu)
 (require 'f)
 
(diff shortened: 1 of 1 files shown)


samhedin (approved):
We've had the issue of test runs failing for no apparent reason before. Simply re-running them can make them succeed. It is, as you say, not related to this PR.
Thanks!

PR #237, opened 2021-03-25
Some assorted minor fixes and cleanup

tarsius:

PR #240, opened 2021-03-31
Another set of assorted changes

tarsius:
After #237 this is the second pull-request with assorted changes that address minor things that I ran to while making major changes (or the third if you count #239).

This should be the last such cleanup pull-request before the real deal is ready.

(I am preparing for the merge of `rust-mode.el` and `rustic.el`, and to make it possible to use `rustic-*.el` (and likewise `rust-*.el`) selectively. I will provide more details once I open that pull-request.)


brotzeit:
Thanks again. I'll check what's wrong with the tests again.

tarsius:
Actually I've been banging on the tests a bit but am not done with that yet. Still you might want to wait a bit to avoid duplicating work.

brotzeit:
Ok, thanks for the info. I know that the tests can be very annoying, so I'm also open to disabling some tests if you think it makes sense.

tarsius:
I have append my build and test fixes.

It is now possible again to use Cask to re-build everything and test using `make clean lisp test`, or to do the same without Cask using `WITHOUT_CASK=true make clean lisp test`. I tend to do the latter because I more strongly dislike Cask every time I encounter it. But everyone else is probably going to use Cask because the other approach assumes that every dependency is installed at `../<name>`, which probably is only true for `borg` and `straight` users, and people who install emacs packages manually.

Regardless of which approach is used this does not report any warning anymore, I fixed them all. When not using Cask, then even the tests are compiled. Many issues were fixed thanks to this. This does not appear to be possible with Cask. Going forward people who edit tests, should manually compile them to avoid introducing a slew of defects again.

I might have fixed the mystery failures along the way. Just in case this resumes again; tests that did more than once randomly fail for me included `rustic-test-babel-crate`, `rustic-test-backtrace` and `rustic-test-save-some-buffers`. So if those start acting up, then we might want to just comment them out. (Though I would of course aprove a proper fix, I don't want to investigate any more time here.)

`rustic-test-babel-crate` (and if that this is disabled, then similar subsequent tests, I believe), wants to unlook the yubikey with my legacy keys, which I always refuse. I couldn't figure out why it would do that and where. I would very much apprechiate if someone else could figure out why it does that and keep it from doing it. Explicitly refusing to unlock seems to have wonders to prevent mystery failures, but now it also succeeds when I let it timeout. All in all I suspect the mystery failures have to do with a race condition between different things that can time out when the key is not unlocked.

When using CI, then the test fail (except on the snapshot) and that is not related to the mystery failures. I know what is going on and it is in part caused by one of my commits. We now need a version of `project` that is more recent than what is released with Emacs 27.1 and earlier. When testing locally that is not a problem because we just install the newer version. On CI it is a problem because CI steps seem to be broken and I don't know how to fix them (nor do I wish to learn how to).

The `Prepare Cask` step appears to fail. It consist of a bunch of commands, but the output that these commands produce are shown in one big block instead of in between the commands that produce it. This is bad for debugging and if the commands can be specified in a way that shows the output in the proper place, then that should be done. Also some output that I would expect to be shown, is missing completely. I would expect `cask install` to output information about the packages that it installs (including `project`), but it does produce no output at all. Maybe that's because it isn't being run at all because the `Prepare Cask` step somehow short-circuits, I don't know. Also I think `cask build` and `cask install` might be specified in the wrong order.

brotzeit:
> rustic-test-babel-crate (and if that this is disabled, then similar subsequent tests, I believe), wants to unlook the yubikey with my legacy keys, which I always refuse. I couldn't figure out why it would do that and where. I would very much apprechiate if someone else could figure out why it does that and keep it from doing it.

We can disable these tests until we fixed it.

I've tried running the tests locally and there still seem to be issues with `rustic-project-buffer-list` and `project-root`. It seems `project-root` was removed in 2015. Do you think we should still use it ?

tarsius:
> I've tried running the tests locally and there still seem to be issues with `rustic-project-buffer-list` and `project-root`. It seems `project-root` was removed in 2015. Do you think we should still use it ?

`project-root` was re-added in 2020 in 5044c19001fe608f2eac621add2e05cbca6c804b, which will be released with Emacs 28.1. I have added `(depends-on "project" "0.3.0")`, which does causes `EMACS=emacs-27.2 cask install` to install the requested version or rather it causes it to report that it is already installed:

```
$ make clean all test
Cleaning...
EMACS=emacs-27.2 cask install
Loading package information... done
Package operations: 13 installs, 0 removals
  - Installing [ 1/13] project (0.3.0)... already present
...
```

I then went to look at that installed `.cask/27.2/elpa/project-0.3.0/project.el` to see whether it really contained the definition of `project-root` only to find out that `project` was not actually installed.

So Cask lied to us about `project 0.3.0` already being installed. I am guessing that it does that for any package that is part of Emacs, even when we request a newer version to be installed than then version that is part of Emacs. Appears to be a Cask bug.

By the way, all the dependencies are wrapped in `(development ...)` which seems to be wrong to me (this are run-time dependencies), for all the other dependencies it seems to work despite of this. Just in case I moved all the `depends-on` to the top-level and also bumped `project` to the very latest version. Didn't appear to make a difference.

I'll report the Cask bug.

tarsius:
That Cask bug has been fixed, so the tests might succeed now.

brotzeit:
> project-root was re-added in 2020 in 5044c19001fe608f2eac621add2e05cbca6c804b, which will be released with Emacs 28.1.

Ah right, I'm currently using an older version.

> That Cask bug has been fixed, so the tests might succeed now.

Nice, never would have hoped this gets merged any time soon given the commit history. Thanks @dickmao.

I'm missing the possibility of travis to delete the cache. I was able to fix many failed test runs by just doing this =)

tarsius:
I have rebased on top of the recent changes that are supposed to fix the test failures. Now we have new unrelated failures earlier in earlier steps that should not at all be affected by the changes I am making.

Maybe it is time to just merge this and trust the the test runs that we perform on our local boxes? This is getting very painful.

brotzeit:
Wow, it's green. These tests can be really frustrating.

PR #242, opened 2021-04-08
fix auto format on remote buffers

### LINKED ISSUE #241: Can't format remote buffers
If I open a TRAMP buffer to a remote `.rs` file, `rustic-format-file` fails with an error like:

```
Error: file `/ssh:<remote>:<path>/src/main.rs` does not exist
```

`rustic-format-buffer` works fine. 

Unsure why, but `rustic-before-save-hook` uses `rustic-format-file` instead of `rustic-format-buffer`. This leads to many annoying popups on remote buffers, since I have formatting configured to run on each save.

A quick solution is to replace `rustic-format-file`with `rustic-format-buffer` in `rustic-before-save-hook`.

dangirsh:
Fixes #241

### DIFF
--- a/rustic.el
+++ b/rustic.el
@@ -1189,17 +1189,20 @@ This handles multi-line comments with a * prefix on each line."
 (defun rustic-before-save-hook ()
   "Don't throw error if rustfmt isn't installed, as it makes saving impossible."
   (when (and (rustic-format-on-save-p) (not (rustic-compilation-process-live t)))
-    (condition-case nil
-        (progn
-          (rustic-format-file)
-          (sit-for 0.1))
-      (error nil))))
+    (if (file-remote-p (buffer-file-name))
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks for opening the pull request. But I just realized we assume that rustfmt is installed locally. Do you think we should rather try to run rustfmt on the remote host ?

dangirsh:
Yes, I noticed that too. I had to disable auto-formatting, since my
channels / architectures on the local and remote don't match.

brotzeit:
FTR we need to use something like `shell-command-to-string` in order to run rustfmt remotely.

brotzeit:
Fixed in https://github.com/brotzeit/rustic/commit/979a3e0034f9ffd773b0b33ff2332f859f5538c8

PR #244, opened 2021-04-12
Revert "Search for Cargo.lock instead of Cargo.toml"

brotzeit:
@hniksic does this work for you ?

### DIFF
--- a/rustic-common.el
+++ b/rustic-common.el
@@ -63,7 +63,7 @@ non-nil."
 If NODEFAULT is t, return nil instead of `default-directory' if directory is
 not in a rust project."
   (let ((dir (locate-dominating-file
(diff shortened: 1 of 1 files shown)


hniksic:
Hi, thanks for looking into this. What would be the best way for me to test the fix - can I update the `rustic` package from Emacs?

brotzeit:
Depends on how you update your packages. For testing purposes, I would simply put the updated function in your config after you required rustic. Or simply eval this function in your running emacs instance.

hniksic:
Thanks, this seems to work much better - `C-c C-c C-k` now runs `cargo check` on the concrete package, not on the whole tree!

Also, `C-c C-c C-c` runs the test at point after building just the current package, which makes it much much faster.

PR #257, opened 2021-06-30
use project-roots instead of project-root (fixes #256)

### LINKED ISSUE #256: uses obsolete project-root function
When I run `rustic-cargo-build` (`C-c C-c C-b`) I get the following error:
```
rustic-project-buffer-list: Symbol’s function definition is void: project-root
```
this appears to be because `project-root` is considered obsolete:
https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/project.el#L235

I noticed that [this commit](https://github.com/brotzeit/rustic/commit/61c37a8dd0ea561ba2641f656cb9a51963af58ad) removed this dependency even though this function is still in use.

How do I get `rustic-cargo-build` to work again?

iitalics:
`project-root` is obsolete and does not exist in some versions ofEmacs*. However it is the same as just `(car (project-roots ..))` so the fix is simple.

* My build: `GNU Emacs 27.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.27, cairo version 1.17.4) of 2021-03-26`

### DIFF
--- a/rustic-rustfmt.el
+++ b/rustic-rustfmt.el
@@ -8,7 +8,7 @@
 
 (require 'rustic-cargo)
 
-(declare-function project-root "project")
+(declare-function project-roots "project")
 
 ;;; Options
 
@@ -226,7 +226,7 @@ This is basically a wrapper around `project--buffer-list'."
     (if (fboundp 'project--buffer-list)
         (project--buffer-list pr)
       ;; Like the above function but for releases before Emacs 28.
-      (let ((root (project-root pr))
+      (let ((root (car (project-roots pr)))
             bufs)
         (dolist (buf (buffer-list))
           (let ((filename (or (buffer-file-name buf)


samhedin:
Related: https://github.com/brotzeit/rustic/pull/232
Emacs 28 says
```
project-roots is a compiled function defined in project.el.

This function is obsolete since 0.3.0; use project-root instead.
```
Like in the created PR, maybe we could use the different functions depending on which version of project.el the user is running?

iitalics:
Ok this is confusing to me because I'm not really knowledgeable on Emacs 27 vs 28 but it looks like 28 is not readily available everywhere. I will probably switch to it but it seems like support for prior releases are important considering the Emacs homepage doesn't provide downloads for 28

iitalics:
At the very least there should be a hard dependency specification in rustic if its using a function which does not exist in older releases

samhedin:
Emacs 28 is the current development version! As you say, support for emacs 27 is important but maybe we can use a switch, otherwise, there will be a need for a new PR whenever Emacs 28 is released and everyone's emacs start to complain about `project-roots` being obsolete instead.

brotzeit:
As @samhedin mentioned, we need to check the emacs version.

PR #258, opened 2021-07-04
Implement `rustic-format-region`.

rbartlensky:
This unfortunately only works with rustfmt nightly, but it's still
an interesting function to have at your disposal.


brotzeit:
Thanks for the pull request!

What do you think about using `(interactive "r")` ? It would then look similar to `rustic-playpen`.

```
(defun rustic-playpen (begin end)
  "Create a shareable URL for the contents of the current region,
src-block or buffer on the Rust playpen."
  (interactive "r")
  (let (data)
    (cond
     ((region-active-p)
      (setq data (buffer-substring begin end)))
...
```

rustic-rustfmt.el
```diff
@@ -188,6 +196,29 @@ and it's `cdr' is a list of arguments."
...
+(defun rustic-format-region ()
+  "Format the current active region using rustfmt.
+
+This operation requires a nightly version of rustfmt.
+"
+  (interactive)
+  (unless (or (eq major-mode 'rustic-mode)
+              (eq major-mode 'rustic-macro-expansion-mode))
+    (error "Not a rustic-mode buffer."))
+  (let ((file (buffer-file-name (current-buffer)))
```
brotzeit, line 209: Maybe it makes sense to use a variable for the current buffer since we use it twice.

rustic-rustfmt.el
```diff
@@ -8,7 +8,9 @@
 
 (require 'rustic-cargo)
 
-(declare-function project-root "project")
+(if (version<= emacs-version "28.0")
+    (declare-function project-roots "project")
+  (declare-function project-root "project"))
```
brotzeit, line 13: Do you think it makes sense to write a simple test that only checks if this is working. This change have been a little annoying #256

rbartlensky: Not sure how I would test this. Do you have any ideas?

rustic-rustfmt.el
rbartlensky: good point!

rbartlensky:
@brotzeit I attempted to address all your comments, although my elisp isn't that great, so I might need another look over this.

rustic-rustfmt.el
```diff
@@ -197,27 +197,36 @@ and it's `cdr' is a list of arguments."
...
-     :command
-     (append (list rustic-cargo-bin "+nightly" "fmt" "--")
-             (rustic-compute-rustfmt-file-lines-args file
-                                                     start
-                                                     (+ start len))))))
+  (let* ((buf (current-buffer))
+         (file (buffer-file-name buf))
+         (r-begin)
+         (r-end))
+    (cond ((region-active-p)
```
brotzeit, line 213: Does the `interactive r` not check if the region is active ?

rustic-rustfmt.el
```diff
@@ -197,27 +197,36 @@ and it's `cdr' is a list of arguments."
...
+          (len (- (count-lines r-begin r-end) 1)))
+      (rustic-compilation-process-live t)
+      (rustic-format-start-process
+       'rustic-format-file-sentinel
+       :buffer buf
+       :command
+       (append (list rustic-cargo-bin "+nightly" "fmt" "--")
+               (rustic-compute-rustfmt-file-lines-args file
+                                                       start
+                                                       (+ start len)))))))
```
brotzeit, line 229: Maybe I'm missing something, but isn't it possible to use the arguments begin and end here ?

rustic-rustfmt.el
brotzeit: No, never mind. Maybe we add something later.

rustic-rustfmt.el
```diff
@@ -188,6 +196,38 @@ and it's `cdr' is a list of arguments."
...
+           (setq r-begin (region-beginning))
+           (setq r-end (region-end))))
+    (let ((start (+ 1 (count-lines 1 r-begin)))
+          (len (- (count-lines r-begin r-end) 1)))
+      (rustic-compilation-process-live t)
+      (rustic-format-start-process
+       'rustic-format-file-sentinel
+       :buffer buf
+       :command
+       (append (list rustic-cargo-bin "+nightly" "fmt" "--")
```
brotzeit, line 226: So this feature is only available on nightly or why do you pin the toolchain ? I guess nowadays lots of people are using stable instead, so we should cover the case when there's no nightly available.

rustic-rustfmt.el
```diff
@@ -188,6 +196,38 @@ and it's `cdr' is a list of arguments."
...
+"
+  (interactive "r")
+  (unless (or (eq major-mode 'rustic-mode)
+              (eq major-mode 'rustic-macro-expansion-mode))
+    (error "Not a rustic-mode buffer."))
+  (let* ((buf (current-buffer))
+         (file (buffer-file-name buf))
+         (r-begin)
+         (r-end))
+    (cond ((region-active-p)
```
brotzeit, line 213: Maybe we should just do something like if region is active -> format region, if not format buffer ?

rustic-rustfmt.el
rbartlensky: I wasn't aware `(interactive "r")` gives you begin and end :sweat_smile: , but now I know. I removed the unnecessary code. However, in this particular case, rustfmt asks you for lines, while begin and end are in bytes (AFAICT).

rustic-rustfmt.el
rbartlensky: Yep, it only works with rustfmt nightly. It looks like they are close to stabilizing (https://github.com/rust-lang/rustfmt/issues/3397)! I'll try to handle this more gracefully

rustic-rustfmt.el
rbartlensky: Sounds good

rustic-rustfmt.el
brotzeit: > However, in this particular case, rustfmt asks you for lines, while begin and end are in bytes (AFAICT).

I didn't notice. Then maybe `(interactive "r")` isn't the right solution. `line-number-at-pos` could help here.

rbartlensky:
@brotzeit thanks again for all your input. I believe I have addressed all your comments :)

rustic-rustfmt.el
brotzeit: But maybe we could only use `end`...

Just do what you think makes sense ;)

rbartlensky: I think I found a good middle-ground solution. Check out the last commit.

rbartlensky:
@brotzeit I think this is ready

brotzeit:
Oh, I missed this pull request, sorry...

Thanks again!

Issue #259, opened 2021-07-05
Changing rust-analyzer path by following README didn't work


veer66:
I use Emacs 27.2 on Windows 10. I tried to change the path to rust-analyzer by follwing README, which is:

> Change rust-analyzer path.
>
> (setq lsp-rust-analyzer-server-command '("~/.cargo/bin/rust-analyzer"))

So I add this line below to my .emacs.

```
(setq lsp-rust-analyzer-server-command '("c:/Users/veer6/pkg/rust-analyzer.exe"))
```
I didn't work. Rustic called rls instead.

However, when I change that line to:

```
(setq rustic-analyzer-command '("c:/Users/veer6/pkg/rust-analyzer.exe"))
```

Now it works.

brotzeit:
Sorry I don't use windows so I don't really think of those small differences. I'm actually surprised the package actually works on windows =)

I would appreciate a pull request that add instructions for windows.

veer66:
I tested on Fedora 34. Rustic + rust-analyer worked. Then I did this.

* I moved rust-analyzer from mv $HOME/.local/bin/rust-analyzer to $HOME/pkg. So it wasn't in $PATH. 
* And then I inserted this to .emacs:

```
(setq lsp-rust-analyzer-server-command '("/home/vee/pkg/rust-analyzer"))
```

* I closed Emacs.
* I opened Emacs again. I opened cli.rs. 
* Rustic used rls instead of rust-analyzer. So it didn't work.
* I changes lsp-rust-analyzer-server-command in .emacs to rustic-analyzer-command. And then it works.

So it is the same on Fedora 34.

brotzeit:
The variable `lsp-rust-analyzer-server-command` gets the value of `rustic-analyzer-command` when you use rustic. Since rustic also provides eglot support this was necessary.

veer66:
> I know it can be confusing so it's best you stick to the rustic readme instead of the lsp-mode docs until you got it working.

Rustic README is this file https://github.com/brotzeit/rustic/blob/master/README.md, isn't it?

> The variable lsp-rust-analyzer-server-command gets the value of rustic-analyzer-command when you use rustic. Since rustic also provides eglot support this was necessary.

I cannot find rustic-analyzer-command in https://github.com/brotzeit/rustic/blob/master/README.md

PR #262, opened 2021-07-08
Use rustic-analyzer-command

veer66:
To fix issue #259, I suppose rust-analyzer's path should be to set to rustic-analyzer-command instead of lsp-rust-analyzer-server-command.


brotzeit:
Thanks! I didn't notice =)

PR #266, opened 2021-08-09
ob-rustic : run code snippets with no main()

flintforge:
Hi there,
this wraps the code block with `main()`
eg: ` fn main(){<body>}` when there's none,
like rust-mode does, allowing more concise snippets.
p.s: the main function detection should be made more robust: by searching for it after rust-fmt

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -266,7 +266,7 @@ kill the running process."
              (dir (setq rustic-babel-dir (expand-file-name project)))
              (wrapped-body (if (string-match-p "fn main()" body)
                                body
-							 (concat "fn main() {\n" body "\n}")))
+                             (concat "fn main() {\n" body "\n}")))
              (main (expand-file-name "main.rs" (concat dir "/src"))))
         (make-directory (file-name-directory main) t)
         (rustic-babel-cargo-toml dir params)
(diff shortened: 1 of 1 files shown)


brotzeit:
Sorry for the late reply.

```
(wrapped-body (if (string-match-p "fn main()" body)
```

I think this regex search can slow things down. I know that some users make excessive use of org-babel and if they use hundreds of lines in babel blocks I think this can be annoying.

I know this would be very convenient and as far as I remember I also considered this when I added the babel code, but I think the potential downsides are not worth it.

Maybe we should rather provide an option for the babel block so you can use a template like:

```
#+BEGIN_SRC rust :wrap t

#+END_SRC
```

flintforge:
Running a regexp without wildcard when interactively executing  a block through babel is barely perceptible.
Ob-C has the option `:main` to tell if it has to check for the its presence of main and eventually add one  (most babel modes do)
Which template would the `:wrap` option select ? Would this provide something that `:noweb` doesn't ?

flintforge:
PR updated to insert the main block as follows:
```
fn main() {
    let result = {
        #{script}
    };
    println!("{:?}", result);
}
```
The usage of ob-rust stays limited to  `:results ouput`.

brotzeit:
> Which template would the :wrap option select ? Would this provide something that :noweb doesn't ?

Something like your example. I haven't used `:noweb`.

> Ob-C has the option :main to tell if it has to check for the its presence of main and eventually add one (most babel modes do)

Ok, I didn't know but certainly sounds reasonable. Maybe we should also do it this way ? We can also add an option to make it the default.

PR #272, opened 2021-08-21
Made cargo-run-current-test choose closest of mod or fn to run.

renatoathaydes:
Previously, this function would always pick the previous fn to run,
even when the current point was a module or inside a module. The
function could be from a previous module, which didn't make sense.
Running all tests in a module is a common thing to do.
This change allows users to do that.
If they are within a fn, the fn only is run.
If they are not within a fn but there's a surrounding mod, then the
whole mod is run.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -140,8 +140,8 @@ When calling this function from `rustic-popup-mode', always use the value of
   "Run 'cargo test' for the test near point."
   (interactive)
   (rustic-compilation-process-live)
-  (-if-let (func-name (rustic-cargo--get-current-fn-fullname))
-      (let* ((command (list rustic-cargo-bin "test" func-name))
+  (-if-let (test-to-run (rustic-cargo--get-test-target))
+      (let* ((command (list rustic-cargo-bin "test" test-to-run))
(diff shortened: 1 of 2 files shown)


rustic-cargo.el
```diff
@@ -154,26 +154,34 @@ When calling this function from `rustic-popup-mode', always use the value of
...
 
 (defun rustic-cargo--get-current-mod ()
-  "Return mod name around point or nil."
+  "Return cons with location and mod name around point or nil."
   (save-excursion
-    (when (search-backward-regexp rustic-cargo-mod-regexp nil t)
-      (match-string 1))))
+    (progn
+      (goto-char (line-end-position))
+      (let ((location (search-backward-regexp rustic-cargo-mod-regexp nil t)))
```
brotzeit, line 174: Maybe use `when-let` ?

renatoathaydes: This is my first elisp code changes ever :) but the language looks pretty cool, I will have a look at improving that.

renatoathaydes: Done. That looks quite good!

brotzeit:
Sorry for the delay and thanks!

brotzeit (approved):

PR #279, opened 2021-10-04
Babel: Toolchain support via `:nightly`

### LINKED ISSUE #277: Specify `toolchain` in `org-babel` snippets
Thank you for Rustic, and especially its amazing support for Org Babel. 

Currently it's possible to specify dependencies (and their features) like so:
```org
#+begin_src rust :crates '((serde . "1.0") (serde_json . "1.0")) :features '((serde . ("derive")))
```
It would be handy if it were possible to also specify `:toolchain nightly` (or similar), so that nightly-gated features could be used in Org code snippets. 

Perhaps it would just be a matter of updating these lines to account for toolchain?

https://github.com/brotzeit/rustic/blob/804ebfe0295a6bf37870e06f84a8d35f55c9f1a6/rustic-babel.el#L55-L58

https://github.com/brotzeit/rustic/blob/804ebfe0295a6bf37870e06f84a8d35f55c9f1a6/rustic-babel.el#L96-L98

fosskers:
This PR allows the user to specify the toolchain via a new `:toolchain` keyword when writing Org snippets.

**Note:** This PR was built off #278, so that should be merged first.

Closes #277

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -50,11 +50,14 @@
 
 (defvar rustic-babel-spinner nil)
 
-(defun rustic-babel-eval (dir)
+(defun rustic-babel-eval (dir toolchain-kw)
   "Start a rust babel compilation process in directory DIR."
   (let* ((err-buff (get-buffer-create rustic-babel-compilation-buffer-name))
          (default-directory dir)
-         (params '("cargo" "build" "--quiet"))
+         (toolchain (cond ((eq toolchain-kw 'nightly) "+nightly")
+                          ((eq toolchain-kw 'beta) "+beta")
+                          (t "+stable")))
+         (params (list "cargo" toolchain "build" "--quiet"))
          (inhibit-read-only t))
     (rustic-compilation-setup-buffer err-buff dir 'rustic-compilation-mode)
     (when rustic-babel-display-compilation-buffer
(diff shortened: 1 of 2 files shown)


brotzeit (approved):

rustic-babel.el
```diff
@@ -50,11 +50,14 @@
...
 
-(defun rustic-babel-eval (dir)
+(defun rustic-babel-eval (dir toolchain-kw)
   "Start a rust babel compilation process in directory DIR."
   (let* ((err-buff (get-buffer-create rustic-babel-compilation-buffer-name))
          (default-directory dir)
-         (params '("cargo" "build" "--quiet"))
+         (toolchain (cond ((eq toolchain-kw 'nightly) "+nightly")
+                          ((eq toolchain-kw 'beta) "+beta")
+                          (t "+stable")))
```
lroolle, line 59: This only allows `nightly`/`beta`/`stable` channels by keyword.

fosskers: As opposed to?

brotzeit: https://rust-lang.github.io/rustup/concepts/toolchains.html#toolchain-specification

fosskers: Ah I see the issue. We could expand the accepted format of `toolchain-kw` to allow for more free-form strings, and pass those as-is (or parse them?) down to the `cargo` line.

brotzeit: Does https://github.com/brotzeit/rustic/commit/d71fbf5f3a39c3f126dc1ae3cafabd16c3160bc0 work for you ?

fosskers: That should do the trick.

fosskers:
That use-case never occurred to me so I wasn't aware that `+stable`, etc., wouldn't be accepted. I suppose this should be considered a bug.

PR #294, opened 2021-10-17
allow user to choose workspace or crate as default for compilation

### LINKED ISSUE #236: Cargo commands run on the whole worskpace
I am working on a large Rust code which is internally divided into crates, and the whole repo is a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) with its own `Cargo.toml`. This allows some flexibility: when working on a particular component, you can build (or check or test) only that crate, but when you want to check how this affects the whole system, you can `cargo test` at top-level.

The issue I am having is that rustic seems to auto-detect the top-level workspace as the "project" and runs commands like <kbd>C-c C-c C-k</kbd> or <kbd>C-c C-c C-c</kbd> on the whole workspace. This is an order of magnitude slower than necessary, and also has issues such as showing warnings or even errors from the crates I'm not interested in. When I was using ordinary rust-mode I would just run `cargo check` from `M-x compile` which would run `cargo check` in the current buffer's directory, and therefore checked only the current crate.

Is there a way to get similar behavior with rustic's compilation features, which are otherwise quite nice?

brotzeit:


ntc2:
Various issues link here, but this PR has no explanation of what the fix was. It seems to be this:
https://github.com/brotzeit/rustic/blob/2c6b363f9ef53e0fb4a63294d877f574262a5404/rustic.el#L89C1-L97C18

```elisp
(defcustom rustic-compile-directory-method 'rustic-buffer-crate
  "Choose function that returns the directory used when calling
 cargo commands.

If you want to use the workspace you can use `rustic-buffer-workspace'.
Note that there may exist functionality that has higher priority than
this variable."
  :type 'function
  :group 'rustic)
```

Issue #295, opened 2021-10-26
Sending input in Cargo run mode errors with wrong-type-argument


jun0:
When I send input to an inferior process in Cargo run mode, upon hitting RET, I get `(wrong-type-argument number-or-marker-p nil)` in `comint-add-to-input-history`.  It seems that `comint-send-input` doesn't work unless comint variables are initialized by a call to the `comint-mode` function.  Additionally, `rustic-cargo-run-mode` calls `kill-all-local-variables`, which undoes all initializations in `comint-mode`, so we somehow need to arrange that initialization to happen after `rustic-cargo-run-mode`.

Unfortunately, I don't know enough about Elisp to know how to make that happen, but here's a workaround for the time being.  It sets up an around-advice that initializes `comint-mode` before `rustic-cargo-run-mode`, then saves and restores comint-related local variables during the initialization of `rustic-cargo-run-mode`.

    (defun my-rustic-cargo-run-mode-comint-fix (orig-fun &rest args)
      (comint-mode)
      (let* ((comint-bindings (cl-loop
                               for (sym . val) in (buffer-local-variables)
                               if (string-prefix-p "comint-" (symbol-name sym))
                               collect (cons sym val)))
             (ret-val (apply orig-fun args)))
        (cl-loop for (sym . val) in comint-bindings
                 do (make-local-variable sym)
                    (set sym val))
        ret-val))
    (advice-add 'rustic-cargo-run-mode
                :around
                #'my-rustic-cargo-run-mode-comint-fix)

System Details:
System: macOS X Big Sur
Emacs version: GNU Emacs 27.2 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95)) of 2021-03-28
Rustic version: rustic-20211018.1934 (pulled from melpa)

brotzeit:
I think it makes sense to keep this open since the fix is still optional.

For anybody who is interested -> https://github.com/brotzeit/rustic#run

brotzeit:
@psibi will add some docs on compilation-mode and comint-mode for 'cargo run'. If somebody has issues with polymode please open a new issue.

@jun0 thanks!

PR #297, opened 2021-10-30
Polymode-based comint interaction with cargo run

jun0:
Hi, this is my attempt at addressing #295.  The dependency on polymode is entirely optional; the user needs polymode only if they enable rustic-cargo-run-use-comint.  One minor issue is its reliance on an undocumented variable in polymode, but hopefully that's OK as: a) polymode is rather careful to implement this functionality, so I don't think it'll go away in the future; b) the whole thing can be disabled if anything goes wrong.

Technical details:
rustic-cargo-run-mode is renamed to rustic-cargo-plain-run-mode.  rustic-cargo-comint-run-mode is added anew.  rustic-cargo-run-mode now refers to a wrapper mode that switches between the two depending on the setting.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -24,6 +24,14 @@ If nil then the project is simply created."
   :type 'boolean
   :group 'rustic-cargo)
 
+(defcustom rustic-cargo-run-use-comint nil
+  "If t then interact with programs in `rustic-cargo-run' using
+comint-mode.  This creates a dependency on the polymode package.
+No special configuration of polymode is needed for this to work,
+but you need to install polymode separately."
+  :type 'boolean
+  :group 'rustic-cargo)
+
 (defvar rustic-cargo-outdated-face nil)
 (make-obsolete-variable 'rustic-cargo-outdated-face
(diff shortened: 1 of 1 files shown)


rustic-cargo.el
```diff
@@ -459,19 +467,68 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
   (buffer-disable-undo)
   (setq buffer-read-only nil)
-  (use-local-map comint-mode-map)
-
-  (define-key (current-local-map) (kbd "RET")
-    '(lambda ()
-       (interactive)
-       (when (eq major-mode 'rustic-cargo-run-mode)
-         (process-send-string
-          (get-buffer-process (current-buffer)) "\n")))))
```
brotzeit, line 474: We didn't merge this code to master, which branch did you checkout ?

jun0: This is based on the cargo-run-send-input branch.

brotzeit: Can you change the target branch to master or is there a reason you didn't use master ?

PR #298, opened 2021-11-01
Cargo run polymode

jun0:
Attempt at addressing #295, take 2, this time targeting master.  Sorry about the clumsiness; I'm not an engineer, and not very used to collaborating on github.

(Rest of the message is same as before.)
The dependency on polymode is entirely optional; the user needs polymode only if they enable rustic-cargo-run-use-comint. One minor issue is its reliance on an undocumented variable in polymode, but hopefully that's OK as: a) polymode is rather careful to implement this functionality, so I don't think it'll go away in the future; b) the whole thing can be disabled if anything goes wrong.

Technical details:
rustic-cargo-run-mode is renamed to rustic-cargo-plain-run-mode. rustic-cargo-comint-run-mode is added anew. rustic-cargo-run-mode now refers to a wrapper mode that switches between the two depending on the setting.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -469,7 +469,13 @@ If running with prefix command `C-u', read whole command from minibuffer."
 (defun rustic-cargo-run-mode ()
   (interactive)
   (if rustic-cargo-run-use-comint
-      (rustic-cargo-comint-run-mode)
+      ;; rustic-cargo-comint-run-mode toggles the mode; we want to
+      ;; always enable.
+      (unless (and (boundp 'polymode-mode)
+                   polymode-mode
+                   (memq major-mode '(rustic-cargo-plain-run-mode
+                                      comint-mode)))
+        (rustic-cargo-comint-run-mode))
     (rustic-cargo-plain-run-mode)))
 
 (define-derived-mode rustic-cargo-plain-run-mode rustic-compilation-mode "Cargo run"
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -482,8 +482,10 @@ If running with prefix command `C-u', read whole command from minibuffer."
   "Mode for 'cargo run' that derives from `rustic-compilation-mode'.
 
 To send input to the compiled program, use
-`rustic-compile-send-input'.  Alternatively, consider enabling
-`rustic-cargo-run-use-comint' (which see)."
+`rustic-compile-send-input'.  If you set
+`rustic-cargo-run-use-comint' to t, you can also just type in a
+string and hit RET to send it to the program.  The latter
+approach requires installing polymode."
   (buffer-disable-undo)
   (setq buffer-read-only nil)
   (use-local-map comint-mode-map))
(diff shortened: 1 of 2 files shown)


brotzeit:
I get `Debugger entered--Lisp error: (void-variable poly-rustic-cargo-compilation-hostmode)`.

rustic-cargo.el
```diff
@@ -458,13 +466,69 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
+    (define-innermode poly-rustic-cargo-comint-innermode
+      :mode 'comint-mode
+      :head-matcher "^ *Running `.+`$"
+      :head-mode 'host
+      :tail-matcher "\\'"
+      :tail-mode 'host)
+    (define-polymode rustic-cargo-comint-run-mode
+      :hostmode 'poly-rustic-cargo-compilation-hostmode
+      :innermodes '(poly-rustic-cargo-comint-innermode)
+      :switch-buffer-functions '(poly-rustic-cargo-comint-switch-buffer-hook)
```
brotzeit, line 512: I think it could be problematic to run these polymode commands every time `rustic-cargo-comint-run-mode` gets called.

Maybe we should check if the polymode modes are already defined.

rustic-cargo.el
```diff
@@ -458,13 +466,69 @@ If running with prefix command `C-u', read whole command from minibuffer."
...
   (setq buffer-read-only nil)
   (use-local-map comint-mode-map))
 
+(defun rustic-cargo-comint-run-mode ()
+  "Mode for 'cargo run' that combines `rustic-compilation-mode' with `comint-mode',
+the former for highlighting and interacting with compiler errors,
+and the latter for interacting with the compiled program."
+  ;; First time around, define the mode and invoke it.  Next time, the
+  ;; symbol will have been overwritten so this runs only once.
+  (require 'polymode)
```
brotzeit, line 499: Maybe we should check if polymode can be required. If not, we can inform the user about the missing package with a message instead of an elisp error.

rustic-cargo.el
jun0: That's already taken care of.  This code runs only once, when `rustic-cargo-comint-run-mode` is called for the first time.  That symbol gets overwritten with `define-polymode`, so from the second time on, the already-defined mode is called.

jun0:
> I get `Debugger entered--Lisp error: (void-variable poly-rustic-cargo-compilation-hostmode)`.

That's strange.  When/where does this happen?

rustic-cargo.el
jun0: Good point.  Fix added.

brotzeit:
I couldn't get this to work, but I can look at it later(it's not that bad even if I'm not the only one, as long as it's optional). Can you add some documentation to the readme so users know about this feature ?

jun0:
> I couldn't get this to work, but I can look at it later(it's not that bad even if I'm not the only one, as long as it's optional). 

OK.  I might also be able to look into it if you can tell me steps to reproduce.

> Can you add some documentation to the readme so users know about this feature ?

Done.

brotzeit:
Thanks!

PR #299, opened 2021-11-03
Remove rustic-top-item-beg-re

rhdxmr:
Use rust-top-item-beg-re instead when hitting C-M-a

I found C-M-a not working with `unsafe fn` and figured out `rust-top-item-beg-re` is overridden by `rustic-top-item-beg-re`


brotzeit:
So the rust-mode version works fine ? Thanks for the contribution.

rhdxmr:
Yes, currently I am using emacs 28 and the latest rustic is working fine with emacs 28.

Issue #291, opened 2021-10-16
Ensure we didn't remove tests that are not present in rust-mode


brotzeit:
It's possible that we've changed the tests that were once pulled from rust-mode during the last years. We removed those tests in #290, maybe there were changes in rustic that are worth restoring in rustic or rust-mode.

brotzeit:
@tarsius it seems you've applied some changes to test files that we removed. Do you think it makes sense to merge those changes into rust-mode ?

tarsius:
Where did I do that?

I probably won't be able to help you with this and similar decision. I don't remember such details as it has been such a long time and because I ended up not using rust (for the time being at least) I am also not particularly motivated to dig in again at this time.

But point me to the changes you are talking about and I will have a look. Who knows, maybe it rings a bell after all.

brotzeit:
It doesn't look important, but thanks for the quick reply. I'll keep it open, maybe I'm motivated to take a look some day =)

And hopefully you get a good reason to take a look at rust again ;)

tarsius:
If you are talking about the `tests` branch in my fork, then I know *nothing* about it. You wrote that change, I forked your repository, you deleted the branch from your repository, and I never got around to pruning old branches.

In fact I would like prune these branches now. So please take a look at all the branches in my fork and move those that you would like to keep over to your repository and then tell me that you are done with that.

brotzeit:
> If you are talking about the tests branch in my fork, then I know nothing about it. 

I mean changes that were merged and afterwards removed.

It seems you already deleted the fork ? I can't find it under your repositories.

I only know that you were working on a transient pr. But I don't know which other feature branches you have. It would certainly be interesting to see that. And I especially would like to have the transient branch. I just noticed that @sehnsucht13 also started working on it two years ago #55

tarsius:
> It seems you already deleted the fork ? I can't find it under your repositories.

It's at https://github.com/tarsiiformes/rustic.

> I only know that you were working on a transient pr.

I actually just started wrapping that up. Would you be interested in that?

> But I don't know which other feature branches you have.

I don't have any. Its just very old branches you authored. There's close to zero chance you would want to recover those, but because of my confusion about what test-changes you were talking about, I though I would mention them.

tarsius:
[...]

PR #301, opened 2021-11-04
Rewrite rustic-cargo using transient

tarsius:
As mentioned in #291.

----

Changes in behavior:

- The popup is closed after invoking a suffix.

- When the prefix is invoked outside any rust project then abort
  instead of prompting for a rust project.

- To abort the popup one has to press "C-g" instead of "q".

- The rustic-popup-commands option that controlled what suffix
  commands are shown in the popup no longer exists.  To remove a
  command one now has to use something like:

    (transient-remove-suffix 'rustic-cargo-build
                             'rustic-cargo-clippy)

- By default it is no longer possible to move around in the popup
  buffer to position the cursor on an action.  But this can be enabled
  for all transients using:

    (setq transient-enable-popup-navigation t)

- Pressing "RET" with point on a suffix in the popup buffer no longer
  invokes the "default action" in all cases.  Instead it invokes the
  selected suffix command just like pressing the respective key would.

  - Previously it was only possible to set the backtrace, compilation
    and test arguments by navigating to the respective line and
    pressing "RET" to invoke the "default action".  If popup
    navigation is enabled, then this is still possible but one can now
    also just press the key that is shown next to the variable(s).

  - Previously it was possible to show the arguments that are
    available for a command by navigating to it and then pressing
    "RET" (or "h").  That is not possible anymore, see next item.

- The ability to show the arguments supported by a command has been
  removed.  This did only show the arguments but did not let the user
  select any of them.  I believe how this was supposed to be used is
  that one would then move to the "Compilation arguments" line and
  enter arguments there.  But it seems that not every command
  respected those arguments and it is likely that not every command
  supports the same set of arguments.  It seems this was just an
  unfinished proof-of-concept.

  Instead the various commands should be defined as prefix commands as
  well, like Magit does it.  For example one of the suffix commands
  available from the `magit-dispatch` transient prefix command is
  `magit-commit`.  That in turn is also a transient prefix command,
  and when it is invoked as a suffix of `magit-dispatch`, then it is
  considered a "sub-prefix".

  I have changed one command, `rustic-cargo-run`, accordingly, but
  without actually listing the relevant arguments.  I cannot do that
  because I don`t use cargo/rust and therefor don't know which
  arguments would make sense.  Someone else will have to do that and
  also define all the other sub-prefixes.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -31,6 +31,11 @@ but you need to install polymode separately."
   :type 'boolean
   :group 'rustic-cargo)
 
+(defcustom rustic-cargo-always-prompt-for-arguments nil
+  "TODO"
+  :type 'boolean
+  :group 'rustic-cargo)
+
 (defvar rustic-cargo-outdated-face nil)
 (make-obsolete-variable 'rustic-cargo-outdated-face
                         "use the face `rustic-cargo-outdated' instead."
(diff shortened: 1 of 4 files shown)


tarsius:
Closing because I won't be working on it, but of course if someone wants to do it, they can clone my fork and take it from there.

Issue #267, opened 2021-08-09
ob-rustic evaluation with cargo script ?


flintforge:
Hi there.

I'm still a noob with cargo but I would like to suggest a way to evaluate code blocks with `cargo script`. 
Here is why:
The two main use cases I see for org-babel are:
1) to test code snippets and to provide code examples
2) in literate programming, when weaving programs using `:noweb` or similar methods.

As we always need to include or import sections of code but also to isolate fragments for testing purpose, eventually the first case get folded into the second one. But the correctness of a section of code isn't necessarily the same as an entire program, right ? The first example that comes to mind being an unused variable declared in one section, but actually used further (e.g. in an other code block).

My suggestion is that we should have two forms of evaluations: 
- for a complete and correct program or program example: going through the standard build procedure. 
- for a more "relaxed" form, eventually evaluating through `cargo script`

Before going through a complex tangle process, this would allow partial evaluation of single blocks while preserving the ability to program with :noweb. What do you think ?

brotzeit:
`:noweb` is for tangling like #122 ?

And you like t o use `cargo script `for stuff like:

```
fn main() {
    let result = {
        #{script}
    };
    println!("{:?}", result);
}
```

so you can use it like 

`$ cargo script -t grabbag -e "mem::size_of::<Box<Read>>()"
16`

but pass `#{script}` as an org-babel argument ?

brotzeit:
Can you maybe provide an example ? I'm not sure I understand.

flintforge:
The example from the cargo-script documentation gives the idea, however I had a better experience with older releases of cargo-script. Both the repos and the cargo release looks presently in a bad shape :/

1) the idea is to evaluate any code statement without a `fn main(){}` wrapping. This is what PR #266 does.
for instance
```
#+begin_src rust
vec![1, 2, 3].iter().map(|&x| x + 1).collect::<Vec<_>>()
#+end_src

#+RESULTS:
: [2, 3, 4]
```

2) The compiler would still complain about unused variables ; That's why I suggest we 
- use cargo-script (or the fork [cargo-eval](https://github.com/reitermarkus/cargo-eval)) or 
- add an `#[allow(dead_code)]` directive in rustic-babel according to some additional predicate

3) In the weaving process

```
#+name: variables
#+begin_src rust
// declare variables here
#+end_src
 
#+begin_src rust :noweb yes
<<variables>>
// make use or variables here
#+end_src
```
local independent evaluations of both code blocks should be correct,
[...]

PR #302, opened 2021-11-05
rustic-babel: support wrap BODY in a fn main if none exists(#267)

lroolle:
#### :main

Auto wrap whole block body in a "fn main" function call if none exists.

Since this is very handy in most code snippets, so the default value is "yes".
"no" if you don't want this feature(for example if you don't want regex search slow things down).

You can also set a default value by:
``` elisp
;; By setq this default to `nil`, you'll have to explict set params to ":main yes" in each block
(setq rustic-babel-auto-wrap-main nil)
```

```
#+begin_src rust
let x = vec![1, 2, 3].iter().map(|&x| x + 1).collect::<Vec<_>>();
println!("{:?}", x);
#+end_src

#+results:
: [2, 3, 4]
```

NOTE: the regex match would unable to /understand/ `fn main()` within a raw string.

```
#+begin_src rust :exports both :main yes
// won't work
let r = "
fn main() {
  let x = "within raw";
}
";

// but this works
printlnt!("fn main{{}}")
#+end_src
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@
             - [:features](#features)
             - [:paths](#paths)
             - [:toolchain](#toolchain)
+            - [:main](#main)
     - [Spinner](#spinner)
     - [inline-documentation](#inline-documentation)
         - [Prequisites](#prequisites)
@@ -79,7 +80,7 @@ The other files provide functionality that is similar to some of the features
 of rustic, however can be considered light-weight compared to some rustic's
 functionality.
 
-The shared functions and options exist as aliases in the rust-mode and 
+The shared functions and options exist as aliases in the rust-mode and
 rustic namespace for backwards compatability reasons(rustic has been a fork).
 
 ## Known issues
@@ -499,7 +500,7 @@ then a string, instead of a list, will also be accepted:
 ```
 #+BEGIN_SRC rust :crates '((tokio . 1.0)) :features '((tokio . ("rt-multi-thread" "time")))
   extern crate tokio;
-  
+
   fn main() {
       tokio::runtime::Runtime::new()
           .unwrap()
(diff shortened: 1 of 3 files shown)


brotzeit:
Thanks for the contribution, with tests :+1: 

But I think emacs users are more used to "t" instead of "yes".

Do you also want to add a `defcustom` that allows the user to set this as a default ? So if the user always wants to wrap the block he doesn't has to pass it as parameter. But we can also open a new issue instead,

lroolle:
> But I think emacs users are more used to "t" instead of "yes".
> 
> Do you also want to add a `defcustom` that allows the user to set this as a default ? So if the user always wants to wrap the block he doesn't has to pass it as parameter. But we can also open a new issue instead.

Good idea, I'll try!

brotzeit:
The parameter should have higher precedence than the `defcustom`.

lroolle:
> The parameter should have higher precedence than the `defcustom`.

Yes of course!

I've add the custom: `rustic-babel-auto-wrap-main` which is boolean, however,  I found the "yes" and "no" parameter would be more reasonable and easy to understood for people who set `:main yes or no`, after ob-C named this, and in elisp you the false value have to set 'nil is not that *explit* I think...

brotzeit:
I think you're right.

lroolle:
@brotzeit 
image

I'm not sure why this failed in the workflow, but ok to run manually(ert run)?

image

brotzeit:
Unfortunately running the tests in ci isn't very reliable.

brotzeit:
Can you do me a favor and reopen the pr with a feature branch. I don't think that's the problem, but who knows ;)

And it's also common practice to use a feature branch instead of master.

README.md
```diff
@@ -541,6 +542,45 @@ fn main() {
...
+#+begin_src rust :exports both :main yes
+// won't work
+let r = "
+fn main() {
+  let x = "within raw";
+}
+";
+
+// but this works
+printlnt!("fn main{{}}")
```
brotzeit, line 580: I think that's not something anybody uses. Do you really think we need this ?

rustic-babel.el
```diff
@@ -26,6 +26,11 @@
   :type 'boolean
   :group 'rustic-babel)
 
+(defcustom rustic-babel-auto-wrap-main t
+  "Whether to auto wrap BODY in a \"fn main\" function call if none exists."
```
brotzeit, line 30: I think you only use upper case letters when it's an argument for a function or am I missing something ?

rustic-babel.el
```diff
@@ -259,6 +264,12 @@ directory DIR."
           (insert s)
           (insert dependencies))))))
 
+(defun rustic-babel-ensure-main-wrap (body)
+  "Wrap BODY in a \"fn main\" function call if none exists."
```
brotzeit, line 268: I would instead use single quotes.

README.md
lroolle: Yes, this is useless...

rustic-babel.el
lroolle: > Can you do me a favor and reopen the pr with a feature branch. I don't think that's the problem, but who knows ;)
> 

Yes, but another PR merge to master or some branch other?

PR #304, opened 2021-11-05
rustic-babel: support auto wrap src block body in a `fn main` if none exists

### LINKED ISSUE #267: ob-rustic evaluation with cargo script ?

lroolle:
close #267(Maybe)

related: #266

#### :main

Auto wrap whole block body in a `fn main` function call if none exists.

Since this is very handy in most code snippets, so the default value is `yes`.
`no` if you don't want this feature(for example, you don't want regex search slow things down).

You can also set a default value by:
``` elisp
;; By setq this default to `nil`, you'll have to explict set params to ":main yes" in each block
(setq rustic-babel-auto-wrap-main nil)
```

```
#+begin_src rust :main yes
let x = vec![1, 2, 3].iter().map(|&x| x + 1).collect::<Vec<_>>();
println!("{:?}", x);
#+end_src

#+results:
: [2, 3, 4]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@
             - [:features](#features)
             - [:paths](#paths)
             - [:toolchain](#toolchain)
+            - [:main](#main)
     - [Spinner](#spinner)
     - [inline-documentation](#inline-documentation)
         - [Prequisites](#prequisites)
@@ -79,7 +80,7 @@ The other files provide functionality that is similar to some of the features
 of rustic, however can be considered light-weight compared to some rustic's
 functionality.
 
-The shared functions and options exist as aliases in the rust-mode and 
+The shared functions and options exist as aliases in the rust-mode and
 rustic namespace for backwards compatability reasons(rustic has been a fork).
 
 ## Known issues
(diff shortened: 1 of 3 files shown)


lroolle:
image

Finally, by rm the following lines, I've got the tests passed:

```elisp
  (let* ((string "let x = \"fn main(){}\";")
         (buf (rustic-test-get-babel-block string)))
    (with-current-buffer buf
      (rustic-test-babel-execute-block buf)
      (should (eq (rustic-test-babel-check-results buf) nil))))
```

Not quite know where's the problem.

brotzeit:
> Not quite know where's the problem.

Yeah, sometimes this fails like magic =)

Thanks for your work!

PR #305, opened 2021-11-06
provide :include to share code between different babel blocks

### LINKED ISSUE #122: Can I share functions between multiple source blocks?
So, I just wanted to ask if I can call a function defined in one org-babel source block to another which are defined in the same sessions. This wasn't working with ob-rust I was trying out earlier, because it didn't allow use of external crates without a main function in that block, but having main function in there meant it couldn't be called from other blocks.

Alternatively, is there a way to add tests? I think I can hack something up if there are tests, although it would be much better to be able to share functions.

brotzeit:


lroolle:
👍  useful feature +1

Issue #322, opened 2021-11-20
auto reload rust-analyzer when Cargo.toml has changed


brotzeit:
https://github.com/hlissner/doom-emacs/issues/5630

I think it makes more sense to add this to rustic instead of lsp-mode.

- [ ] lsp-mode
- [ ] eglot

brotzeit:
This is actually a lsp client issue. I'll keep it open until we know it's working for both clients.

brotzeit:
I've tested it with lsp-mode and eglot and both seem to work ? I think we should discuss this in lsp-mode/eglot, we have better chances to get some help there.

jadestrong:
<img width="688" alt="image" src="https://user-images.githubusercontent.com/18475217/153796201-85c2ca10-29e1-4deb-a860-94bc1826c95e.png">

I tried it on Linux and Windows. Their `file-notify--library` is`inotify` and `w32notify` respectively, which can work normally, but `kqueue` is used on my Mac computer, so it still can't work normally here.

brotzeit:
I didn't even know about this =)

Sorry, but I can't help here.

PR #323, opened 2021-11-20
enable automatic reloading of rust-analyzer when Cargo.toml changed

brotzeit:
#322

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,10 @@ info](https://github.com/rust-lang/rustfmt/issues/4454).
 Disable LSP support by setting `rustic-lsp-client` to nil. You have to
 restart emacs when you switch lsp clients.
 
+Changes to Cargo.toml triggers reloading of rust-analyzer
+automatically after saving. If you want to disable this feature you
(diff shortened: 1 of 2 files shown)


brotzeit:
I guess we also have to check if there's a running server.

flodiebold:
rust-analyzer can already reload the workspace if it is notified about changes in Cargo.toml. If that somehow doesn't work, I think it'd be a much better idea to fix that instead of putting in a workaround that results in worse user experience.

brotzeit:
> rust-analyzer can already reload the workspace if it is notified about changes in Cargo.toml. 

And how does this currently happen when using lsp-mode ?

flodiebold:
It is supposed to happen through the LSP file watching mechanism; rust-analyzer registers file watches including Cargo.toml, and when it receives a `didChangeWatchedFiles` notification about these files, it reloads the workspace.

brotzeit:
Makes sense. I'll ensure that it really doesn't work and open an lsp-mode issue instead. Thanks for your help!

PR #327, opened 2021-11-26
rustic-buffer-workspace should return the directory instead of Cargo.toml

rich-murphey:
rustic-compile-directory-method expects a function that returns the path to a directory.

Setting `(setq rustic-compile-directory-method 'rustic-buffer-workspace)` seems to have an issue when 
rustic-buffer-workspace returns /root/of/workspace/Cargo.toml.   In this case, rustic-cargo-check fails with the message:
```
apply: Setting current directory: Not a directory, /root/of/workspace/Cargo.toml
```

This patch removes the /Cargo.toml from the path, so that rustic-buffer-workspace returns the parent dir.

### DIFF
--- a/rustic.el
+++ b/rustic.el
@@ -72,7 +72,8 @@
           (error "`cargo locate-project' returned %s status: %s" ret (buffer-string)))
         (goto-char 0)
         (let* ((output (json-read))
-               (dir (cdr (assoc-string "root" output))))
+               (dir (replace-regexp-in-string "/Cargo.toml$" ""
+                                              (cdr (assoc-string "root" output)))))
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks for the fix. I think we can use `file-name-directory` instead.

rich-murphey:
Sounds great.  Thanks so much!

brotzeit:
@rich-murphey can you apply the change ?

brotzeit:
Thanks :+1:

rich-murphey:
I tested file-name-directory successfully.   Thanks again for rustic!

PR #350, opened 2021-12-31
Fix macro expanding, remove incorrect nightly check

puuuuh:
Fix macro expanding, remove incorrect nightly check in rustic-format-region, update syntax and idents in rustic-macro-expansion-mode

### DIFF
--- a/rustic-lsp.el
+++ b/rustic-lsp.el
@@ -134,12 +134,22 @@ with `lsp-rust-switch-server'."
 
 (define-derived-mode rustic-macro-expansion-mode special-mode "Rust"
   :group 'rustic
-  :syntax-table rustic-mode-syntax-table
+
+  :syntax-table rust-mode-syntax-table
+
+  ;; Syntax
+  (setq-local syntax-propertize-function #'rust-syntax-propertize)
+
+  ;; Indentation
+  (setq-local indent-line-function 'rust-mode-indent-line)
+
   ;; Fonts
-  (setq-local font-lock-defaults '(rust-font-lock-keywords
-                                   nil nil nil nil
-                                   (font-lock-syntactic-face-function
-                                    . rust-mode-syntactic-face-function))))
+  (setq-local font-lock-defaults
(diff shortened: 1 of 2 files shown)


rustic-lsp.el
```diff
@@ -152,18 +162,13 @@ with `lsp-rust-switch-server'."
       (let ((inhibit-read-only t))
         (erase-buffer)
         ;; wrap expanded macro in a main function so we can run rustfmt
-        (insert "fn main()")
+        (insert "fn main() {")
```
brotzeit, line 165: I guess they have changed the output.

rustic-rustfmt.el
```diff
@@ -274,6 +294,20 @@ were issues when using stdin for formatting."
                                :buffer (current-buffer)
                                :stdin (buffer-string)))
 
+(defun rustic-format-macro-buffer ()
+  "Format the current buffer with macro output using rustfmt.
+
+Provide optional argument NO-STDIN for `rustic-before-save-hook' since there
+were issues when using stdin for formatting."
```
brotzeit, line 301: This sentence doesn't fit here.

rustic-lsp.el
puuuuh: Attribute macroses can expand to one or more statements and cant be formatted with old method

rustic-rustfmt.el
```diff
@@ -261,10 +281,7 @@ This operation requires a nightly version of rustfmt.
 
 ;;;###autoload
 (defun rustic-format-buffer ()
-  "Format the current buffer using rustfmt.
-
-Provide optional argument NO-STDIN for `rustic-before-save-hook' since there
-were issues when using stdin for formatting."
```
brotzeit, line 267: It seems the line wasn't removed when we changed this function in the past. Thanks for fixing it.

rustic-lsp.el
brotzeit: I see.

PR #351, opened 2022-01-03
Add rustic-cargo-test-mode-map

psibi:
This overrides the keybinding `g` to rustic-cargo-test-run which IMO
is more appropriate than just compiling in the rustic-cargo-test-mode


brotzeit:
Thanks!

PR #354, opened 2022-01-07
Improve cargo run functionality when used without comint

psibi:
This improves the plain cargo run functionality:

- Makes cargo run buffer non editable. The original behaviour is
retained when comint is enabled.
- Introduces rustic-run-arguments similar to rustic-test-arguments
- Binds the key 'g' for doing re-run in the cargo run buffer

Note: This functionality needs to be tested by people who use comint
functionality with rustic-mode (Although I don't understand it's
usecase since Rust doesn't have a REPL workflow)

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -486,6 +486,9 @@ As a byproduct, you can run any shell command in your project like `pwd'"
   (setq command (read-from-minibuffer "Command to execute: " (car compile-history) nil nil 'compile-history))
   (rustic-run-cargo-command command (list :mode 'rustic-cargo-run-mode)))
 
+(defvar rustic-run-arguments ""
+  "Holds arguments for 'cargo run', similar to `compilation-arguments`.")
+
 ;;;###autoload
 (defun rustic-cargo-run (&optional arg)
   "Run 'cargo run' for the current project.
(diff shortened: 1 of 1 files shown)


brotzeit:
> Although I don't understand it's
usecase since Rust doesn't have a REPL workflow

I think it's about user input for programs that open a prompt.

psibi:
> I think it's about user input for programs that open a prompt.

Ah, I see. :-) I will try to document it in my upcoming MR then.

psibi:
> What is the intention ?

This is my usecase: I have a non interactive console program which I'm protyping and want to do `cargo run` quite a bit. Just pressing `g` from the `cargo run` buffer is quite handy and this PR implements that kind of workflow.

> It seems many users that make use of this feature want to be able to send input. Do you think we should only allow editing in rustic-cargo-comint-run-mode ?

Agree. I have to play around with it to understand this usecase better. Worst case, I think an workflow facilitating both would be to do add a two key binding to `rustic-cargo-run-rerun`. 

I think as a next step, I can enable comint and `polymode` package and understand the workflow better and probably update the PR accordingly. WDYT ?

psibi:
> What is the intention ?

This is my usecase: I have a non interactive console program which I'm protyping and want to do `cargo run` quite a bit. Just pressing `g` from the `cargo run` buffer is quite handy and this PR implements that kind of workflow.

> It seems many users that make use of this feature want to be able to send input. Do you think we should only allow editing in rustic-cargo-comint-run-mode ?

Agree. I have to play around with it to understand this usecase better. Worst case, I think an workflow facilitating both would be to do add a two key binding to `rustic-cargo-run-rerun`. 

I think as a next step, I can enable comint and `polymode` package and understand the workflow better and probably update the PR accordingly. WDYT ?

brotzeit:
> Just pressing g from the cargo run buffer is quite handy and this PR implements that kind of workflow.

I could have figured this out myself =)

> Worst case, I think an workflow facilitating both would be to do add a two key binding to rustic-cargo-run-rerun.

IMO if users want a comint mode they should use a different mode. Most users don't need it and prefer being able to use shortcuts like in compilation mode.

So we probably should make it clear in the docs that the default is NOT comint and if they want to use it they have to use polymode with the rustic comint mode. I think it would be good to extend the "Cargo/Run" section in the readme and add headings for each mode and explain the differences. If you want to do it would be nice, but you don't have to.

psibi:
> IMO if users want a comint mode they should use a different mode. Most users don't need it and prefer being able to use shortcuts like in compilation mode.

Agree! I will see if I can create something along those lines in the coming days. Thanks for the answers!

PR #359, opened 2022-01-09
Refactor and improve the ability to do cargo run commands

psibi:
Summary of the changes:

- We have two modes now: rustic-cargo-run-mode and rustic-cargo-run-comint-mode. The first one is primarily meant for non interactive usage and the comint mode is used for passing data to it.
- Remove the dependency on polymode. The whole thing felt hacky IMO. If you want to have highlighting on compile errors, the recommendation is to just use `rustic-compile` before going to the run phase.
- We just pass the data directly in the comint buffer now instead of using `rustic-compile-send-input`. I have removed that function now as it's not required anymore.


psibi:
@brotzeit It seems strange that the test suite is failing with 27.2 and passing with snapshot. And the test seems unrelated with this MR. Do you know what's going wrong ?

brotzeit:
Sometimes you have to rerun the tests in github.

brotzeit:
> Remove the dependency on polymode. The whole thing felt hacky IMO. If you want to have highlighting on compile errors, the recommendation is to just use rustic-compile before going to the run phase.
We just pass the data directly in the comint buffer now instead of using rustic-compile-send-input. I have removed that function now as it's not required anymore.

I know it's not perfect, but I agreed on adding this feature so I actually don't want to remove it since somebody already took the time to implement it(would be rude IMO). And since this is only an optional feature we can document it properly and users can decide if they want to use it(I wouldn't make this a default). Maybe somebody finds a less hackish way in the future. 

Give some time for review and to get an opinion on this, I rarely use cargo run.

psibi:
@brotzeit Fair point, what do you think about three modes then:

- rustic-cargo-run-mode
- rustic-cargo-run-comint-mode
- rustic-cargo-plain-run-mode - We recover the functionality under the polymode.

README.md
```diff
@@ -151,7 +151,6 @@ Commands:
 
 - `rustic-compile`            compile project using `rustic-compile-command`
 - `rustic-recompile`          recompile using `compilation-arguments`
-- `rustic-compile-send-input` send string to process of current buffer
```
brotzeit, line 154: I'd like to keep this function since it allows you to send input to the process even if you are using the compilation mode.

psibi: @brotzeit Thanks for the review. I have restored the function and have also restored the polymode based functionality. I have added the details in the README on how it differss.

brotzeit:
Ok let's give this a try(I didn't do an in-depth review). I'll add some basic tests. Next time it would be better if you could split cleanup changes and new features, makes reviewing easier.

However I have one last request. Can you please move the comint related stuff into a new file rustic-comint.el. I would do it myself but I don't want to put my name on the changes.

psibi:
@brotzeit Thanks, will make sure to give a smaller PR's in future. :-)

psibi:
Thanks for merging it. Any idea on what's the source of the test failures are ? From a quick look it doesn't seem to be `run` related.

brotzeit:
https://github.com/brotzeit/rustic/commit/2c3a8f185a918203d404eb78c2c4031495c562ac maybe ? And sometimes tests in github actions fail for no reason. You can always run tests locally to check if they are really failing. Otherwise you have to rerun the tests in github until they go green ;)

PR #363, opened 2022-01-13
rustic-doc: Support Eglot, neither eglot nor LSP-mode required

samhedin:
Support for eglot in addition to lsp-mode. If neither exists, don't crash and just work somewhat anyway.


samhedin:
@brotzeit I'm having an issue where if I remove lsp-mode and activate eglot, then run `rustic-doc-mode`, it will crash and complain that LSP-mode is missing. However, if I open `rustic-doc.el` and evaluate it, everything works fine. Do you have any idea of what can be going on? I'm trying to understand if it's my emacs config that's doing it or something in my code.

samhedin:
@brotzeit if you try this with eglot, are you seeing the same issue as I'm having? I'm not sure if it's my emacs that's loading something wrong when I start it.

brotzeit:
I'll take a look.

brotzeit:
I gave it another try and I had issues with eglot so I tried lsp-mode.

First I had to manually set `rustic-doc-current-project` in `rustic-doc-convert-current-package` so the process starts.

Then I ran into a new issue, here is the output of 

```
[WARNING] Duplicate identifier 'blanket-implementations' at input line 645 column 3305
Error running filter /home/fredbert/.local/bin/rustic-doc-filter.lua:
Error during function call: /home/fredbert/.local/bin/rustic-doc-filter.lua:58: attempt to index a nil value (field 'List')
stack traceback:
	/home/fredbert/.local/bin/rustic-doc-filter.lua:58: in function </home/fredbert/.local/bin/rustic-doc-filter.lua:53>
	[C]: in ?
	[C]: in function 'pandoc.walk_block'
	/home/fredbert/.local/bin/rustic-doc-filter.lua:164: in function 'Pandoc'
stack traceback:
	[C]: in function 'pandoc.walk_block'
    /home/fredbert/.local/bin/rustic-doc-filter.lua:164: in function 'Pandoc'
```

It seems there are some functions that need an update. Usually I would say let's add some tests to make things easier. But I know it's not that easy when lsp is involved...

samhedin:
Hmm, frustrating. Yes, it's time to add some tests. Most functionality does not rely on LSP, luckily.

brotzeit:
I have a way better feeling when adding tests, I break things too often =)

brotzeit:
Let's give it a try.

PR #369, opened 2022-01-22
Rustic-doc-tests

samhedin:
About time I learned how these things work


brotzeit:
Just a few words...

Don't be too pedantic about tests. I don't expect "clean" code for tests(because I didn't do it from the beginning). Just pick some test cases you think make sense. And add the number of tests you think are appropriate, you implemented the feature.

samhedin:
@brotzeit yeah, no worries. I've wanted to do this for a long time (I even had a branch in the old repo), I've just been lazy.
But a question, because one of the issues you were having in the other thread is the same as something I've noticed all around the place and for a very long time.
Try this:
1. Restart emacs and open a rust project
2. Run `rustic-doc-mode`
3. Run `rustic-doc--update-current-project` (not interactive, so eval it)
Now that should set `rustic-doc-current-project` to the opened rust project, but it doesn't. What you need to do is open `rustic-doc.el`, evaluate the buffer, then it will work when you run `rustic-doc--update-current-project`.
Do you understand what's going on? Why is it not being loaded properly? I used to think there was something wrong with my emacs config that made this behavior, but now that you're seeing it that seems unlikely.

brotzeit:
I've played around with the code a little and noticed 2 things that might play a role. We don't require rustic-doc.el and 

```diff
modified   rustic-doc.el
@@ -205,7 +205,7 @@ it doesn't manage to find what you're looking for, try `rustic-doc-dumb-search'.
 
 (defun rustic-doc--update-current-project ()
   "Update `rustic-doc-current-project' if editing a rust file, otherwise leave it."
-  (when (and lsp-mode
+  (when (and (featurep 'lsp-mode)
              (derived-mode-p 'rust-mode 'rustic-mode))
     (setq rustic-doc-current-project (lsp-workspace-root))))
```

samhedin:
Sounds reasonable.  Should `rustic-doc` be required inside the `define-minor-mode` macro?

brotzeit:
Ideally we require it where the other files are required. I also don't see another option.

samhedin:
Okay, that was frustrating. That was a lot of debugging back and forth just to find that the test was the major source of errors.
Some nice things were discovered though: Debian and its derivatives have some *really* old versions of fd-find, ripgrep and pandoc, and this can cause issues. Ripgrep and fd-find can easily be installed through cargo, but pandoc needs some intervention from the user.

rustic-doc is now required as you specify so that should help out. I'll add more tests as we go, but for now this should be fine to merge.

rustic-doc.el
```diff
@@ -101,16 +101,20 @@ The function should take search-dir and search-term as arguments."
   :type 'function
   :group 'rustic-doc)
 
+
+
 (defun rustic-doc--install-resources ()
   "Install or update the rustic-doc resources."
+  (executable-find "chmod")
```
brotzeit, line 108: What's the purpose of this line ?

rustic-doc.el
```diff
@@ -311,9 +331,11 @@ NO-DL is primarily used for development of the filters."
                              (lambda (_p)
                                (message "Finished converting docs for std"))
                              "std")
+  (sleep-for 3)
```
brotzeit, line 334: Maybe wait for the process to finish instead ? Or what's reason that we have to wait here ?

rustic-doc.el
```diff
@@ -311,9 +331,11 @@ NO-DL is primarily used for development of the filters."
...
   (rustic-doc-convert-current-package))
 
 (defun rustic-doc--start-process (name program finish-func &rest program-args)
+  (print (format "rustic-doc--start-process. name: %s program:  %s" name program))
```
brotzeit, line 338: I think `message` is the common way to inform the user about progress.

brotzeit:
> Okay, that was frustrating. That was a lot of debugging back and forth just to find that the test was the major source of errors.
Some nice things were discovered though

I know writing tests in elisp can be tedious.

rustic-doc.el
samhedin: Also a relic from debugging. I should start taggin these as todo or something to remind myself to remove them.

rustic-doc.el
samhedin: This was to prevent minibuffer messages to instantly be overwritten by other minibuffer messages. I didn't see a way to avoid it at the time (even if I agreed that adding sleeps here and there is not great), but it's changed now.

rustic-doc.el
brotzeit: No problem, that's what reviews are for.

rustic-doc.el
brotzeit: Sometimes sleeping is necessary, unfortunately.

brotzeit (approved):
Merge when you're ready.

PR #371, opened 2022-01-27
Universal argument for cargo clean command

psibi:
This makes it quite similar to the other functions like build, run commands present in the rustic mode.

My personal use case is to pass `-p package_name` as I'm hitting this bug way too often currently: https://github.com/rust-lang/rust/issues/84970


samhedin (approved):
Seems to work as advertised. Thank you!

PR #373, opened 2022-01-28
Insert command into the compilation buffer

### LINKED ISSUE #372: Print the command executed in the rustic-compilation buffer
Now now when we do rustic build etc, we get a new `rustic-compilation` buffer which looks like this:

```
   Compiling my-package v0.9.0
...
...
```

Instead having something like this would be better IMO:

```
  cargo build my-super-package --some-option  

   Compiling my-package v0.9.0
...
...
```

I can volunteer to implement this if others think that this isn't a bad idea. :-)

psibi:
Fixes #372

Sample `rustic-cargo-clean`:

``` shellsession
cargo clean

rust-compilation finished at Fri Jan 28 08:51:40
```

And when we pass custom parameter:

``` shellsession
cargo clean -p package

rust-compilation finished at Fri Jan 28 08:54:06
```

### DIFF
--- a/rustic-compile.el
+++ b/rustic-compile.el
@@ -288,9 +288,8 @@ ARGS is a plist that affects how the process is run.
     (unless (plist-get args :no-display)
       (funcall rustic-compile-display-method buf))
     (with-current-buffer buf
-      (read-only-mode -1)
-      (insert (format "%s \n" (s-join " "  command)))
-      (read-only-mode 1)
+      (let ((inhibit-read-only t))
+        (insert (format "%s \n" (s-join " "  command))))
       (rustic-make-process :name process
                            :buffer buf
                            :command command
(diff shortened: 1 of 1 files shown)


rustic-compile.el
```diff
@@ -287,6 +288,9 @@ ARGS is a plist that affects how the process is run.
     (unless (plist-get args :no-display)
       (funcall rustic-compile-display-method buf))
     (with-current-buffer buf
+      (read-only-mode -1)
+      (insert (format "%s \n" (s-join " "  command)))
+      (read-only-mode 1)
```
brotzeit, line 293: The idiomatic way of doing this would be using a `let` binding with `(inhibit-read-only nil)`

psibi: Thanks, updated!

brotzeit:
:+1:

PR #376, opened 2022-01-31
Fix bugs in rustic-cargo-outdated-mode

psibi:
Fixes two bugs:
- Fixes a bug when dependencies are specified as git repositories:

The `cargo outdated --depth 1` doesn't always gives the package from line number 2. For a sample project I'm having this is the output it gives:

``` shellsession
❯ cargo outdated --depth 1
    Updating git repository `https://github.com/snoyberg/async-refresh-rs`
Name           Project  Compat   Latest   Kind    Platform
----           -------  ------   ------   ----    --------
anyhow         1.0.40   1.0.53   1.0.53   Normal  ---
askama         0.10.5   ---      0.11.0   Normal  ---
```

As you can see in the above case, the hardcoding of the value 2 doesn't work. The above scenario, makes cargo outdated stuff to break entirely. Infact, for me the buffer was empty and the sentinel function broke. The fix instead tries to check for `--` 
 character to delimit where the packages are starting from.

- Fix cargo upgrade when you update version to latest version

Right now, rustic-cargo-upgrade-execute only works when we try to upgrade it to a compabible version ie

``` shellsession
Name           Project  Compat   Latest   Kind    Platform
----           -------  ------   ------   ----    --------
anyhow         1.0.40   1.0.53   1.0.53   Normal  ---
```

Trying to upgrade above will work. But this will fail:

``` shellsession
Name           Project  Compat   Latest   Kind    Platform
----           -------  ------   ------   ----    --------
askama         0.10.5   ---      0.11.0   Normal  ---
```

And the reason is that it tries to use this code `(elt crate 2)` for taking the version from. But this makes it to work only when the
`compat` column is already filled. Also because of the way `tabulated-list` works, we just change the buffer when we do `rustic-cargo-mark-upgrade`. This commit fixes it in such a way that it works for all the cases.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -357,30 +357,42 @@ Execute process in PATH."
   (interactive)
   (tabulated-list-put-tag " " t))
 
+(cl-defstruct rustic-crate name version)
+
+(defun rustic-cargo--outdated-make-crate (crate-line)
+  "Create RUSTIC-CRATE struct out of a CRATE-LINE.
+
+The CREATE-LINE is a single line from the `rustic-cargo-oudated-buffer-name'"
+  (make-rustic-crate :name (nth 1 crate-line) :version (nth 2 crate-line)))
+
 ;;;###autoload
 (defun rustic-cargo-upgrade-execute ()
   "Perform marked menu actions."
   (interactive)
-  (let (crates)
-    (save-excursion
-      (goto-char (point-min))
-      (while (not (eobp))
-        (let* ((cmd (char-after))
-               (crate (tabulated-list-get-entry (point))))
-          (when (eq cmd ?U)
-            (push crate crates)))
-        (forward-line)))
+  (let ((crates (rustic-cargo--outdated-get-crates (buffer-string))))
     (if crates
-        (let ((msg (format "Upgrade %s ?" (mapconcat #'(lambda (x) (elt x 0)) crates " "))))
+        (let ((msg (format "Upgrade %s ?" (mapconcat #'(lambda (x) (rustic-crate-name x)) crates " "))))
           (when (yes-or-no-p msg)
             (rustic-cargo-upgrade-crates crates)))
       (user-error "No operations specified"))))
+
+(defun rustic-cargo--outdated-get-crates (cargo-outdated-buffer-string)
+  "Return a list of RUSTIC-CRATE which needs to be updated.
+
+ CARGO-OUTDATED-BUFFER-STRING represents the entire buffer of
+`rustic-cargo-oudated-buffer-name'"
+  (let* ((lines (s-lines cargo-outdated-buffer-string))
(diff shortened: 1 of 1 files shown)


rustic-cargo.el
```diff
@@ -349,30 +357,42 @@ Execute process in PATH."
...
+  (let ((crates (rustic-cargo--outdated-get-crates (buffer-string))))
     (if crates
-        (let ((msg (format "Upgrade %s ?" (mapconcat #'(lambda (x) (elt x 0)) crates " "))))
+        (let ((msg (format "Upgrade %s ?" (mapconcat #'(lambda (x) (rustic-crate-name x)) crates " "))))
           (when (yes-or-no-p msg)
             (rustic-cargo-upgrade-crates crates)))
       (user-error "No operations specified"))))
 
+(defun rustic-cargo--outdated-get-crates (cargo-outdated-buffer-string)
+  "Return a list of RUSTIC-CRATE which needs to be updated.
```
brotzeit, line 380: I guess you forgot to update "RUSTIC-CRATE" after you removed the argument.

rustic-cargo.el
```diff
@@ -349,30 +357,42 @@ Execute process in PATH."
   (interactive)
   (tabulated-list-put-tag " " t))
 
+(cl-defstruct rustic-crate name version)
+
+(defun rustic-cargo--outdated-make-crate (crate-line)
+  "Create RUSTIC-CRATE struct out of a CRATE-LINE.
+
+The CREATE-LINE is a single line from the `rustic-cargo-oudated-buffer-name'"
```
brotzeit, line 365: typo -> CREATE-LINE

rustic-cargo.el
```diff
@@ -349,30 +357,42 @@ Execute process in PATH."
   (interactive)
   (tabulated-list-put-tag " " t))
 
+(cl-defstruct rustic-crate name version)
+
+(defun rustic-cargo--outdated-make-crate (crate-line)
+  "Create RUSTIC-CRATE struct out of a CRATE-LINE.
```
brotzeit, line 363: Another upper case RUSTIC-CRATE

rustic-cargo.el
brotzeit: Ok, got it.

rustic-cargo.el
psibi: Fixed, thanks!

rustic-cargo.el
psibi: I made it in backticks. From a quick look into the the EMACS manual, I wasn't able to find/confirm if that's right. But I will further investigate it tomorrow. Thanks!

brotzeit: AFAIK you only make functions arguments in upper case, but I'm not 100% sure.

brotzeit: checkdoc usually knows what's common practice. But I don't like it as part of the test suite.

PR #380, opened 2022-02-05
fix relative path with rustic-buffer-crate

### LINKED ISSUE #370: Errors are not clickable with a workspace
When `mycrate` is in a workspace, compilation (<kbd>C-c C-c C-b</kbd>) links in the error messages change from `src/lib.rs` to `mycrate/src/lib.rs` and `next-error` does not find the correct file to go to.  It would be really nice if this worked (workspaces come rather rapidly for separation of concerns — say a `*-sys` crate and a higher level one developed together).

brotzeit:
close #370


brotzeit:
🤞

brotzeit:
I realized that we already have an advice for `compile-goto-error` so I used this.

PR #381, opened 2022-02-06
cargo expand integration

psibi:


README.md
```diff
@@ -511,6 +511,15 @@ before being used.
 
 ![](https://raw.githubusercontent.com/brotzeit/rustic/master/img/outdated.png)
 
+### Expand
```
brotzeit, line 514: We need to update the toc.

brotzeit:
Can you do me a favor and generate a project instead of checking it into the repo. I just added the test project for some test cases that would require additional functionality for generating projects. In this case we don't need a multi-crate workspace.

psibi:
@brotzeit You mean generate a new cargo project which is not part of workspace ? Or you mean use some elisp code to generate a project ?

I'm fine doing either both with a slight preference to a plain cargo project as I think that makes things simpler.

brotzeit:
We have functionality to generate a project automatically. Something like:

```elisp
(ert-deftest rustic-cargo-expand-test ()
  (let* ((string "fn main() { () }")
         (buf (rustic-test-count-error-helper-new string))
         (default-directory (buffer-file-name buf)))
    (with-current-buffer buf
      (call-interactively 'rustic-cargo-expand)
      (let* ((proc (get-process rustic-expand-process-name))
             (buffer (process-buffer proc)))
        (while (eq (process-status proc) 'run)
          (sit-for 0.01))
        (with-current-buffer buffer
          (let ((buf-string (buffer-substring-no-properties (point-min) (point-max))))
            (should (string-match "^cargo expand" buf-string))))))))
```

psibi:
Seems the tests are failing because there is no cargo-expand in the $PATH. I will try to fix them tomorrow.

psibi:
This MR is ready for review. I have slightly tweaked the cargo expand test to not actually test the buffer contents since it currently errors out in the CI system because of absence of the `cargo-expand` executable. I didn't find any binary being released as part of official repos and doing `cargo install` will add around ~4 to 5 minutes of CI times which seemed unreasonable for a single test.

README.md
psibi: Fixed, thanks!

brotzeit:
LGTM!

Issue #386, opened 2022-02-15
Configurable build arguments for cargo


rwaweber:
Thanks for the project, very handy!

In my experience, I've found that a handful of rust projects often require a specific selection of flags and features when being built on windows.

As a result, I was wondering if you'd be open to a PR that added a `rustic-cargo-build-arguments` option, very similar to these two [settings](https://github.com/brotzeit/rustic/blob/master/rustic-cargo.el#L43-L51).

An initial skim makes it seem like it could require a little fiddling with [`rustic-cargo-build`](https://github.com/brotzeit/rustic/blob/master/rustic-cargo.el#L600-L604).

In the event that I'm overlooking another setting that toggles this feature, happy to close this out!

brotzeit:
Sure :+1:

PR #388, opened 2022-02-16
[cargo] add cargo-build-arguments as variable

rwaweber:
In order to maintain backwards compatibility, keep value empty.

Should close https://github.com/brotzeit/rustic/issues/386.


brotzeit:
Thanks!

PR #400, opened 2022-04-26
Fix `cargo doc` commands

arminfriedl:
`rustic-run-cargo-command` wants each argument as separate string in the
list.

`shell-command` only takes a single string `command` and fails when
providing a list.


brotzeit:
Thanks!

PR #402, opened 2022-05-08
Add `cargo login` command

arminfriedl:
`cargo login <TOKEN>` adds a token for https://crates.io to the local
`$CARGO_HOME/credentials` file. This is then used for `cargo`
commands that somehow interact with crates.io like `cargo publish`.


arminfriedl:
Thought I'd add it since its part of `cargo` not sure if it was intentionally left out of `rustic-cargo` of course

brotzeit:
:+1:

PR #404, opened 2022-05-16
remap recompile keybinding and enable `compilation-start-hook`

CeleritasCelery:
Two changes here:

First is to run `compilation-start-hook` as part of staring a new rustic process. This will let it more closely mirror the functionality of `compile` and work with code that uses that hook (which I often do for setting timers or watchers).

Second we change the hardcoded `g` keybinding for `rustic-recompile` to use the `remap` directive. This will ensure seamless behavior if the user has a different key for recompile (such as when using `evil-mode`).


brotzeit:
Thanks!

PR #406, opened 2022-05-23
Support custom cargo check/run/build/test commands

muqiuhan:
Sometimes we don't just want to run cargo check/run/test/build commands, but also cargo xrun/xcheck/xtest/xbuild, the purpose of this pr is to support customization of these commands


brotzeit:
:+1:

Issue #173, opened 2020-10-22
Is it normal that the *rustic-compilation* buffer is read only?


parvizfarnia:
I'm completely a beginner in using `rustic` and I wasn't sure to write this topic here as an "Issue" or rather post it on a forum because I don't really know whether it is actually a bug. Here is my problem: When I press `C-c C-c C-r` the program is run via `cargo run` and there is a new buffer in Emacs labled: `* rustic-compilation *` where you can view the output of `cargo run`. However it seems to me that unlike a classic terminal you don't have the possibility to provide any input value to the program. Consider the following example (from The Book on Rust web site)

```
use std::io;


fn main() {
    println!("Guess the number!");

    println!("Please input your guess.");

    let mut guess = String::new();

    io::stdin()
        .read_line(&mut guess)
        .expect("Failed to read line");

    println!("You guessed: {}", guess);
}
```

Here the program, waits for the user to enter a value, yet there is no possibility to enter a value in the `* rustic-compilation *` buffer and it seems to me that this buffer is somewhat locked/read-only. Am I missing something here or this is indeed how the buffer has been implemented, that is without any user interaction?

Thanks in advance,

parvizfarnia:
I slightly modified a solution that I found in this thread: [Cargo-process does not accept user input](https://emacs.stackexchange.com/questions/51156/cargo-process-does-not-accept-user-input)

The following (added to my init.el) solved the problem:

```
(defun my-cargo-run ()
  "Build and run Rust code."
  (interactive)
  (rustic-cargo-run)
  (let (
      (orig-win (selected-window))
      (run-win (display-buffer (get-buffer "*rustic-compilation*") nil 'visible))
    )
    (select-window run-win)
    (comint-mode)
    (read-only-mode 0)
  )
)
(define-key rustic-mode-map (kbd "C-*") 'my-cargo-run)

```

brotzeit:
Yeah I just realized that regular compilation-mode also behaves this way. I will change it.

dcguim:
I am still experiencing the same problem in package 20220625.2008. When I execute `rustic-cargo-run` in my rustic session, and attempt to write in the *cargo-run* emacs buffer I get:
```
user-error: No (^error[^:]*:[^
]*
 *--> \([^
]+\):\([0-9]+\):\([0-9]+\) 1 2 3) here
```
The rust program I am executing is exactly the same as above.

dcguim:
[...]

PR #421, opened 2022-07-11
rustic-cargo-comint-run was causing a Lisp error - wrong-type-argumen…

dcguim:
When attempting to run `M-x rustic-cargo-comint-run` for rust code reading from stdin I bumped into this [issue](https://github.com/brotzeit/rustic/issues/173#issuecomment-1179407436). The wrong-type-argument error was captured in this piece of code in the `rustic-cargo-run-get-relative-example-name()` function, the often nil `buffer-file-name` was causing the error.
 ```elisp
(if buffer-project-root
              (split-string (file-relative-name buffer-file-name buffer-project-root) "/") nil)))
```
The correct buffer file name is the name of the directory where the current source file sits (called here `buffer-folder`). I have obtained the name for this directory from `(rustic-buffer-crate)`, in the following form:
```elisp
(let* ((buffer-project-root (rustic-buffer-crate))
         (buffer-path-list (split-string buffer-project-root "\\/"))
	 (buffer-folder (nth (- (length buffer-path-list) 2) buffer-path-list)) 
...
```
the last line is simply fetching the last folder in the path retrieved by `(rustic-buffer-crate)`.

### DIFF
--- a/rustic-comint.el
+++ b/rustic-comint.el
@@ -54,7 +54,7 @@ If ARG is not nil, use value as argument and store it in `rustic-run-arguments'.
 When calling this function from `rustic-popup-mode', always use the value of
 `rustic-run-arguments'."
   (interactive "P")
-  (let* ((run-args (rustic--get-run-arguments)))
+  (let ((run-args (rustic--get-run-arguments)))
     (pop-to-buffer-same-window
      (get-buffer-create rustic-run-comint-buffer-name))
     (unless (comint-check-proc (current-buffer))
--- a/rustic-comint.el
+++ b/rustic-comint.el
@@ -54,18 +54,19 @@ If ARG is not nil, use value as argument and store it in `rustic-run-arguments'.
 When calling this function from `rustic-popup-mode', always use the value of
 `rustic-run-arguments'."
   (interactive "P")
-  (pop-to-buffer-same-window
-   (get-buffer-create rustic-run-comint-buffer-name))
-  (unless (comint-check-proc (current-buffer))
+  (let* ((run-args (rustic--get-run-arguments)))
+    (pop-to-buffer-same-window
+     (get-buffer-create rustic-run-comint-buffer-name))
+    (unless (comint-check-proc (current-buffer))
     (rustic--cargo-repl-in-buffer
      (current-buffer)
      (concat "run" (cond
                     (arg
                      (setq rustic-run-comint-arguments
                            (read-from-minibuffer "Cargo run arguments: " rustic-run-comint-arguments)))
-                    ((rustic--get-run-arguments))
+                    (run-args)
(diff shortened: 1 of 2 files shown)


psibi:
@dcguim Do you think changing `rustic-cargo-comint-run` into this will instead fix it:

```
(defun rustic-cargo-comint-run (&optional arg)
  "Run 'cargo run' but for interactive programs.

If ARG is not nil, use value as argument and store it in `rustic-run-arguments'.
When calling this function from `rustic-popup-mode', always use the value of
`rustic-run-arguments'."
  (interactive "P")
  (let* ((run-args (rustic--get-run-arguments)))
    (pop-to-buffer-same-window
     (get-buffer-create rustic-run-comint-buffer-name))
    (unless (comint-check-proc (current-buffer))
    (rustic--cargo-repl-in-buffer
     (current-buffer)
     (concat "run" (cond
                    (arg
                     (setq rustic-run-comint-arguments
                           (read-from-minibuffer "Cargo run arguments: " rustic-run-comint-arguments)))
                    (run-args)
                    (t ""))))
    (rustic-cargo-run-comint-mode))))
```

dcguim:
Thanks, @psibi that works as well.

psibi:
@dcguim Can you update your PR with that ? 

Regarding the changes on `rustic-cargo-run-get-relative-example-name`, I'm not sure they are really required. But probably someone else can confirm that.

dcguim:
Will do. Yeah, I got more stable results when getting path information from `(rustic-buffer-crate)` perhaps we should be more careful with the setting of `buffer-file-name`:

_"""Local in buffer rustic-cargo.el; global value is nil._ 

_Automatically becomes permanently buffer-local when set.
This variable may be risky if used as a file-local variable.
  Probably introduced at or before Emacs version 17."""_

rustic-comint.el
```diff
@@ -54,18 +54,19 @@ If ARG is not nil, use value as argument and store it in `rustic-run-arguments'.
 When calling this function from `rustic-popup-mode', always use the value of
 `rustic-run-arguments'."
   (interactive "P")
-  (pop-to-buffer-same-window
-   (get-buffer-create rustic-run-comint-buffer-name))
-  (unless (comint-check-proc (current-buffer))
+  (let* ((run-args (rustic--get-run-arguments)))
```
psibi, line 57: I think just having `let` shoud do. Sorry, about that.

psibi (approved):
LGTM, apart from a minor change.

Regarding changes on `rustic-cargo-run-get-relative-example-name`, probably open a new PR and we can discuss ? I think finding some examples where it does/doesn't work in the existing case might help things out.

psibi:
Thank you!

Issue #423, opened 2022-08-02
`rust-format-file` hangs on Windows


ptillemans:
On Emacs-28.1 for Windows (emacs-version : Emacs 28.1 (build 2, x86_64-w64-mingw32) of 2022-04-21) when running M-x rust-format-file the command does not terminate until C-c is pressed. 

Some debugging learned that the process-status keeps in the 'run state. I could not repeat it when experimenting running rustfmt with the same arguments in the scratch buffer. 

It smells like rustfmt detects the input file is opened and waiting for input, but it could be another reason why it decides to hang around till I press C-c. (and that does not explain why on Linux in WSL2 it works fine).

brotzeit:
Does `rustic-format-buffer` work ?

ptillemans:
Yes, that works fine. I should have  mentioned that. That's what I found puzzling prompting me to dig a bit deeper but I got stuck because ert-runner does not work properly either, got sidetracked and then my time ran out.

brotzeit:
I just tried it on my windows machine and it works for me. Did you try it in different projects ?

brotzeit:
You could comment the code in `rustic-format-file` step by step and see what's actually causing the problem. I'm sorry but as I can't reproduce I don't know how to help.

ns-mkusper:
Same issue on the same OS and version of Emacs. 

To me it looks like rustfmt processes Emacs processes endlessly run even though  the rustfmt.exe command exits immediately. I believe this is because `rustic-make-process` is setting the process coding system to `'utf-8-emacs-unix` which then causes mismatches for signals on non-unix systems. 

Unless I'm missing something setting this to `default-process-coding-system` should more reliably ensure that the appropriate coding system gets used here across different environments. 

Opened a PR with that change. #426 

@ptillemans Can you try that on your end as well?

mylo-oleary:
Hi, I'm experiencing this issue on Windows as well. The fix PR that was merged only adds to the problem, since when I try to call `rust-cargo-run` (as well as potentially other rustic functions) with the current code (as of the PR), the command seems to run but I get an error saying that the wrong type argument was passed to the set-process-coding-system function. Following this error message, every attempt to call `rust-cargo-run` fails.

[...]

PR #426, opened 2022-08-18
Fix `rustic-format-file` hang on Windows

ns-mkusper:
Setting process coding system to `utf-8-emacs-unix` for rustic processes causes issues when reading process signals on non-unix systems. An example of this is #423 . 

This PR sets the coding system of new rustic processes to the value of `default-process-coding-system` which should more reliably ensure that an appropriate coding system is used across environments.

### DIFF
--- a/rustic-compile.el
+++ b/rustic-compile.el
@@ -238,7 +238,11 @@ Set environment variables for rust process."
       (run-hook-with-args 'compilation-start-hook process)
       (set-process-filter process (plist-get args :filter))
       (set-process-sentinel process (plist-get args :sentinel))
-      (set-process-coding-system process 'utf-8-emacs-unix 'utf-8-emacs-unix)
+      ;; Workaround for rustic-format-file hanging bug on Windows
+      ;; See https://github.com/brotzeit/rustic/issues/423
+      (if (eq system-type 'windows-nt)
+          (set-process-coding-system process default-process-coding-system default-process-coding-system)
+        (set-process-coding-system process 'utf-8-emacs-unix 'utf-8-emacs-unix))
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks for the pr. However this doesn't work for me as `default-process-coding-system` is a cons but we need a symbol. Maybe use `buffer-file-coding-system` instead ? Seems to work.

ns-mkusper:
Ah, I had overlooked that mismatch but you're right. However, when trying `buffer-file-coding-system` or even `(car default-process-coding-system)` and `(cdr default-process-coding-system)` for the decoding and encoding values I still experience the hang in Windows. Only when I set the `ENCODING` argument to `default-process-coding-system` do I see normal performance with `rustic-format-file` which is curious. 

I suspect there's something odd going on with the emacs primitives which are then being called with those arguments that's making the difference but I'll have to take a closer look to confirm exactly why I'm seeing this behavior. Will update once I get a chance to debug further.

brotzeit:
Curious. We could also add a temp fix `(when (eq system-type windows-nt) ...)`. At least rustfmt would finish this way. Looks like this could indeed by an emacs bug so we need a temporary fix anyway...

ns-mkusper:
Unfortunately that too results in hanging for me on Windows 10/11 on `rustic-format-file` . It only seems to go through if I use the `default-process-coding-system` values which as we mentioned is less than ideal since that's a workaround that depends on some undefined behavior. 

I updated the PR for now to at least only update this behavior for Windows.

brotzeit:
Let me check if it works on my windows machine.

brotzeit:
Still doesn't hang for me.

PR #429, opened 2022-08-20
Automatically add missing dependencies

### LINKED ISSUE #428: Add missing dependencies automatically on using `use <pkg>`
Based on https://github.com/brotzeit/rustic/releases/tag/3.3 where the following feature was introduced 

> `rustic-cargo-add-missing-dependencies`: add missing dependencies to Cargo.toml, works with lsp-mode and eglot

Is there a way to make it a variable which if true it'll be added automatically? I think lsp in C adds headers automatically when you use a function from un-included libraries.

Ideally when you do `use <something>;` just on that line. I think. Can't currently remember how lsp provides completions on use. But if we can modify it, the completions could be made with already available crates on top and also crates that can be added to Toml so if they're chosen it activates the add to toml function. Might have to use/update a list of available crates from cargo locally for that though.

If it becomes too complex or not possible (doubt that considering it's emacs) then doing it automatically on save is also fine.

brotzeit:
close #428

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,8 @@ Customization:
 - `rustic-cargo-bin-remote` Path to remote cargo executable
 - `rustic-cargo-build-arguments` default arguments for cargo build
 - `rustic-cargo-check-arguments` default arguments for cargo check
+- `rustic-cargo-auto-add-missing-dependencies` automatically add missing dependencies
+to Cargo.toml by checking new diagnostics for 'unresolved import' errors(currently only works with lsp-mode)
 
 ### Edit
 
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -80,6 +80,14 @@ If nil then the project is simply created."
   :type 'string
   :group 'rustic-cargo)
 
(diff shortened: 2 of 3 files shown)


brotzeit:
I have to find a way to avoid the noise when 'cargo add' is failing when the imported crate doesn't exist.

brotzeit:
One downside:
When there are multiple unresolved imports and one of the crates doesn't exist, no dependencies will be added. We could only avoid this by adding the crates one by one. Don't know if it's worth the overhead.

Atreyagaurav (approved):

Atreyagaurav:
Oh  I need to be one with write access to approve, sorry.

Anyway that downside is ok, since if there is/are unresolved imports we're not going to be able to compile that code anyway, and when it's resolved then all of them will be added to `Cargo.toml` right? Then it's not a problem at all.

brotzeit:
> and when it's resolved then all of them will be added to Cargo.toml right?

As soon all dependencies can be found, they will be added.

PR #434, opened 2022-08-23
uses `process-file` instead of `call-process` to respect remoteness of default directory

### LINKED ISSUE #432: `cargo locate-project` returned status 101
Hi all! I was trying to edit a remote file and invoking `rustic-cargo-run` when I got the above error. Tracing back the error, I located this line of code:

https://github.com/brotzeit/rustic/blob/f57cce4a8faf243b8b6541554d2a9fb134214201/rustic.el#L82

which invokes cargo via `call-process`. I don't believe this command respects the remote status of the `default-directory`, which means if `cargo` is available at both the local machine and tramp, it may not use the right version of cargo. Changing this function to call `cargo` using `process-file`, which does respect `default-directory`, seemed to enable `rustic-cargo-run` on my end to work.

I didn't want to make a PR in case I was missing something in my setup. Thanks so much for your help!

Jemoka:
closes #432

perhaps #433 is a more robust solution (that actually resolves for the remote path), but this stopgap solution worked for me because it would just check for cargo on the remote machine first.

### DIFF
--- a/rustic.el
+++ b/rustic.el
@@ -79,7 +79,7 @@
         (setq-local process-environment env)
         ;; Set PATH so we can find cargo.
         (setq-local exec-path path)
-        (let ((ret (call-process (rustic-cargo-bin) nil (list (current-buffer) nil) nil "locate-project" "--workspace")))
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks!

PR #435, opened 2022-08-27
enable user to configure cargo-build argument when C-u is pressed

taquangtrung:
Hi,

I created a small PR that allows users to configure `rustic-cargo-build-arguments` when running `cargo-build` by pressing the prefix argument key `C-u`.

Can you review and consider merging it?

Thanks!


brotzeit:
Thanks for the pr.

PR #436, opened 2022-08-27
make the prefix argument of cargo-build optional

taquangtrung:
Hi,

I just received a failed CIL build, not sure if that was because of my commit.

But I added a small fix to make the argument of `rustic-cargo-build` optional.
Maybe it can help to resolve the failed CI.


brotzeit:
Oh right, didn't review properly...

PR #445, opened 2022-09-23
add bound check to `rustic-cargo-auto-add-missing-dependencies`

toyboot4e:
Hi :wave: When I set `rustic-load-optional-libraries` to `nil`, I got this error on M-x `rustic-mode`:

> Symbol's value as variable is void: rustic-cargo-auto-add-missing-dependencies

I think it's because `rustic-cargo.el` is not loaded and their custom variables are not set.

This PR fixes it by adding `bound-and-true-p` check.

### DIFF
--- a/rustic.el
+++ b/rustic.el
@@ -152,7 +152,7 @@ this variable."
 \\{rustic-mode-map}"
   :group 'rustic
 
-  (when rustic-cargo-auto-add-missing-dependencies
+  (when (bound-and-true-p rustic-cargo-auto-add-missing-dependencies)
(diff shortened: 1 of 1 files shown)


brotzeit:
Thanks :+1: 

I think the option was supposed to be removed after rust-mode was added as dependency(but we obviously forgot it). Out of curiosity, why do you use rustic if you disable loading all rustic features. Why not just use rust-mode instead ?

toyboot4e:
Right, I would switch back to `rust-mode`, thanks! I forgot why I was using rustic 🤔

PR #453, opened 2022-10-17
rustic-save-some-buffers-advice: respect the `pred' argument of `save-some-buffers'

deftsp:
In some situation like `org-save-all-org-buffers', one might do not want to save the rust  buffer


brotzeit:
Thanks!

PR #457, opened 2022-11-07
Revert "Fix `rustic-format-file` hang on Windows"

ns-mkusper:
Reverts brotzeit/rustic#426

Given that #426 is now known to cause issues ( reported in [issue 423](https://github.com/brotzeit/rustic/issues/423#issuecomment-1302800867) ) without a definite understanding of how this 'fix' worked in the first place reverting it is the best course. If a true fix is found after more investigation I'll be sure to add another PR for that.


brotzeit:
Thanks!

PR #465, opened 2022-11-17
introduce new option that allows reusing args with cargo commands

brotzeit:
@taquangtrung works ?

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -65,6 +65,18 @@ If nil then the project is simply created."
   :type 'boolean
   :group 'rustic-cargo)
 
+(defcustom rustic-cargo-use-last-stored-arguments nil
+  "Always rerun cargo commands with stored arguments.
+
+Example:
+When `rustic-cargo-use-last-stored-arguments' is `nil', then
+rustic-cargo-test will always use `rustic-default-test-arguments'.
+
+If you set it to `t', you can reuse the arguments with `rustic-cargo-test'
+instead of applying the default arguments from `rustic-default-test-arguments'."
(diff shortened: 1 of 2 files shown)


rustic-cargo.el
```diff
@@ -65,6 +65,19 @@ If nil then the project is simply created."
   :type 'boolean
   :group 'rustic-cargo)
 
+(defcustom rustic-cargo-use-last-stored-arguments nil
+  "Always rerun cargo commands with stored arguments.
+
+Example:
+When turned off, running `rustic-cargo-test' with `C-u' will store the
+arguments in `rustic-test-arguments' and allows you to reuse them with
+`rustic-cargo-test-rerun'.
```
taquangtrung, line 74: This description is a bit confusing.

When `rustic-cargo-use-last-stored-arguments` is `nil`, then `rustic-cargo-test` will always use `rustic-default-test-arguments`.

Should the part about `C-u M-x rustic-cargo-test` be described for the function `rustic-cargo-test` but not for this `rustic-cargo-use-last-stored-arguments` variable?

brotzeit: > Should the part about C-u M-x rustic-cargo-test be described for the function rustic-cargo-test but not for this rustic-cargo-use-last-stored-arguments variable?

We should apply this variable also for other cargo commands that have rerun commands. I think it's too much if we add a sentence to every command.

brotzeit: Thanks for the review, I applied your suggestion.

taquangtrung: @brotzeit: thanks for your effort! :+1:

PR #469, opened 2022-12-12
Add way to load per-project rust-analyzer settings.

rbartlensky:
Most rust-analyzer config flags are passed via initializationOptions, which means we can't use eglot's `.dir-locals.el`. Instead, this patch adds support for reading the options from the project's root from a special file called `.rust-analyzer-init.json`.

At the moment only `eglot` is supported.

`rustic` already has support for `detachedFiles` which is merged with the config that is loaded from `.rust-analyzer-init.json`.

I've succesfully tested the following json locally:
```json
// .rust-analyzer-init.json
{
    "cargo": {
        "extraEnv": { "CARGO_TARGET_DIR": "./target-ra" }
    }
}
```

I am opening this ~as a draft PR~ (EDIT: I refreshed the page and accidentally opened this as a regular PR...) to open a discussion about this feature. I worked on a couple of projects that had their own fancy rust-analyzer settings for VS code, and I wanted to have something similar for emacs + eglot + rust-analyzer. First of all, is this the right place to add such support? If not, where? Second of all, does the implementation make sense?

### DIFF
--- a/rustic-lsp.el
+++ b/rustic-lsp.el
@@ -127,18 +127,40 @@ with `lsp-rust-switch-server'."
                                    eglot-server-programs)))))
     (add-to-list 'eglot-server-programs `(rustic-mode . (eglot-rust-analyzer . ,rustic-analyzer-command)))))
 
+(defun rustic-get-string-from-file (file-path)
+  "Return file content as string."
+  (with-temp-buffer
+    (insert-file-contents file-path)
+    (buffer-string)))
+
+(defun rustic-load-rust-analyzer-init ()
+  "Load '.rust-analyzer-init.json' from the project's root"
+  (let ((init-file (concat default-directory ".rust-analyzer-init.json")))
+    (if (file-exists-p init-file)
+        (condition-case err
+            (json-parse-string (rustic-get-string-from-file init-file))
+          (error (message "failed to parse '%S': %S" init-file err) eglot--{}))
+      eglot--{})
+    ))
+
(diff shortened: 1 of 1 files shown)


rbartlensky:
Thanks for the quick reply!

> I never understood why the emacs lsp clients don't support this, I think lsp-bridge works this way. But there's probably a reason for it that I don't know. I'm surprised that this is the first issue I've seen that mentions this.

When I first read about `.dir-locals.el` I thought that's all I had to do to be fair (and I guess that works for most usecases, but not rust-analyzer)

> This should be implemented in eglot/lsp-mode. However if you don't get this merged we can give it a try(since we already have this hack for `:detachedFiles`). Maybe you can open an issue and ask if it would be accepted.

I opened a discussion [here](https://github.com/joaotavora/eglot/discussions/1135) if you're interested. I will keep this PR open for now if you don't mind, at least until some kind of conclusion is reached.

psibi:
> I never understood why the emacs lsp clients don't support this, 

I think this shouldn't be a problem for lsp-mode atleast as it already supports this kind of workflow for one of the language servers: https://emacs-lsp.github.io/lsp-mode/page/lsp-sqls/#sample-configuration

rbartlensky:
I will close this for now. It looks like there is a way to configure eglot's initializeParameters via `.dir-locals`, more info in the discussion. I might open some a PR later to add some docs about this if you think that's worth it

PR #474, opened 2022-12-20
Fixes edge case of running examples and plain binary

psibi:
This also makes sure that cargo-run works in both the workflows: with and without popup/ in plain binary crate as well as examples.

Notes about changes:

- When we invoke rustic popup, the buffer-file-name becomes nil which caues it to break when running examples. So we create a new variable to tracke the source file name now.
- We optimize rustic--get-run-arguments to avoid calling to rustic-cargo-run-get-relative-example-name
- We use rustic-run-arguments as last resort. It's a good fallback to have IMO and it's default value is nil anyway.

### DIFF
--- a/rustic-popup.el
+++ b/rustic-popup.el
@@ -50,6 +50,9 @@ The first element of each list contains a command's binding."
 (defvar rustic-popup-buffer-name "rustic-popup-buffer"
   "Buffer name for rustic popup buffers.")
 
+(defvar rustic--popup-rust-src-name nil
+  "Rust source code file name from which rustic-popup was invoked.")
+
 (defvar rustic-popup-mode-map
   (let ((map (make-sparse-keymap)))
     (define-key map [remap self-insert-command] 'rustic-popup-invoke-popup-action)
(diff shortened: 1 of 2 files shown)


rustic-popup.el
```diff
@@ -50,6 +50,9 @@ The first element of each list contains a command's binding."
 (defvar rustic-popup-buffer-name "rustic-popup-buffer"
   "Buffer name for rustic popup buffers.")
 
+(defvar rustic--popup-rust-src-name nil
```
brotzeit, line 53: Why the extra "-" ?

brotzeit: Does this make the variable local to rustic ? I've only seens this with `rustic-popup--`.

brotzeit: I obviously haven't used this enough :D

brotzeit (approved):

rustic-popup.el
psibi: Yeah, mostly a convention to indicate that it's private. Do you want me to change it to `rustic-popup--` ? No strong opinions from my side.

brotzeit: I think it makes more sense to keep it this way because it's also used in another file.

PR #477, opened 2022-12-21
Improve lsp-mode language server client interaction with cargo fmt

psibi:
Currently, when you invoke rustic-cargo-fmt it takes a quite a bit of time because lsp-mode's rust analyzer connection disconnects and connects with each buffers that are opened in the Emacs buffer. Infact, if you visit your Messages buffer, you would be able to see the logs:

LSP::disconnected
LSP::connected with rust-analyzer

While it is a minor inconvenince for simple projects, for projects with huge number of files - it's causes a big delay.

Looking at the lsp-mode's code, this is the reason:

- When a major mode changes, the lsp-mode disconnects - https://github.com/emacs-lsp/lsp-mode/blob/a2353d0bee8465ce87f7cef1b2ee9b285cb865e2/lsp-mode.el#L4017

And the reason it is getting disconnected is, that rustic-cargo-fmt function doesn't preserve major mode when it is refreshes the text from the content present on the disk.

The fix for it is simple, it makes sure that we don't alter the file modes.

### DIFF
--- a/rustic-rustfmt.el
+++ b/rustic-rustfmt.el
@@ -275,7 +275,7 @@ and it's `cdr' is a list of arguments."
                           (funcall rustic-list-project-buffers-function))))
             (dolist (b buffers)
               (with-current-buffer b
-                (revert-buffer t t)))))
+                (revert-buffer t t t)))))
         (kill-buffer proc-buffer)
         (message "Workspace formatted with cargo-fmt.")))))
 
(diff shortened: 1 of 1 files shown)


brotzeit:
Never realized this. Do you think it makes sense to also apply this fix to the other revert-buffer calls ?

psibi:
> Never realized this. 

I guess I was the only one too annoyed by this that I tried to avoid using rustic provided formatting functionalities. I usually tried to use the rust-analyzer's functionality for formatting buffer - but that didn't work for formatting workspaces. So I finally had enough motivation to track this down. :-)

> Do you think it makes sense to also apply this fix to the other revert-buffer calls ?

Yeah, good call. It's always good to preserve the major mode. I have pushed a fix for the same.

brotzeit (approved):

brotzeit:
> but that didn't work for formatting workspaces

I guess that's just a matter of time.

PR #478, opened 2022-12-22
Improve expand macro and LSP integration

### LINKED ISSUE #473: Improve expand macro and LSP integration
Right now, this functionality is implemented as part of `rustic-analyzer-macro-expand` function. But it has a couple of issues:

- It wraps the expanded macro in a main function so that rustfmt can be run on it. This leads to wrong code in most cases: All the expanded code is present inside the main function, when it is not the case. So probably after we do the formatting step, we should remove the main block. (This leads to a even more confusing situation when we use tokio::main macro. We have two main functions!)
- When we try to do `rustfmt` on the expanded code, it almost always fails for me. These are some of the error messages:

```
error: expected identifier, found keyword `let`
 --> <stdin>:3:9
  |
2 |     let body = async {
  |                ----- `async` blocks are only allowed in Rust 2018 or later
3 |         let Cmd { opt, subcommand }: Cmd = Cmd::parse();
  |         ^^^ expected identifier, found keyword
  |
  = help: pass `--edition 2021` to `rustc`
  = note: for more on editions, read https://doc.rust-lang.org/edition-guide
```

In my opinion, we should to this change:

- Do not try to format the macro as the response from the language server comes in the formatted fashion: https://github.com/rust-lang/rust-analyzer/pull/12014
- Avoid the extra wrapping to please `rustfmt` as it's not required anymore.

I will be happy to provide a patch if there are no objections.

psibi:
We do the following changes:

- Do not try to format the macro as the response from the language server (rust-analzyer) comes in the formatted fashion: https://github.com/rust-lang/rust-analyzer/pull/12014
- Avoid the extra wrapping of main to please rustfmt as it's not required anymore.
- Make rustic-analyzer-macro-expand non interactive as it's not meant to be called interactively. It's called from the lsp-mode's rust client integration.

Ultimately this fixes #473

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,7 +381,7 @@ top of the sideline.
 ##### Macro expansion
 
 `lsp-rust-analyzer-expand-macro` expand macro call at point
-recursively
+recursively.
 
 The results are formatted and highlighted by default, but you can use
 your own function by customizing
--- a/rustic-lsp.el
+++ b/rustic-lsp.el
@@ -167,22 +167,15 @@ with `lsp-rust-switch-server'."
 
 ;;;###autoload
 (defun rustic-analyzer-macro-expand (result)
-  "Default method for displaying macro expansion results."
-  (interactive)
+  "Default method for displaying macro expansion RESULT ."
   (let* ((root (lsp-workspace-root default-directory))
          (buf (get-buffer-create
                (format "*rust-analyzer macro expansion %s*" root))))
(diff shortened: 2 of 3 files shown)


brotzeit:
Can you please also remove `rustic-format-macro-buffer` and `rustic-format-macro-sentinel`.

brotzeit (approved):

PR #479, opened 2022-12-24
Rustic playground improvements

psibi:
This is similar to my rust-mode's PR, but for rustic: https://github.com/rust-lang/rust-mode/pull/480

Additionally this also creates rustic-playpen-buffer similar to the one present in rust-mode.


brotzeit:
I think the name playpen is outdated,somebody mentioned it a while ago. Can you Change ist to playground.

brotzeit:
But it's not important, we can also do it later. I think we are already at least 5 years late =)

brotzeit (approved):

psibi:
> But it's not important, we can also do it later. I think we are already at least 5 years late =)

Sorry, been bit busy with work. I will fix this one and the rust-mode PR by hopefully tomorrow.

brotzeit (approved):

Issue #475, opened 2022-12-20
Provide option to use native emacs treesitter mode instead of rust-mode for emacs29+


brotzeit:
https://git.savannah.gnu.org/cgit/emacs.git/commit/?h=emacs-29&id=cb8ccdd26702606710258e2e08bd1fc35bbc1674

psibi:
How are you planning to integrate this option ? 

Some questions:

- Will a new rustic-mode will be derive on top of the provided tree sitter based mode ? I personally think this is the future, but not sure about others. :-)
- Will this new mode will be on-par with the current existing regex based one ? If not, where do they differ ?

brotzeit:
> Will a new rustic-mode will be derive on top of the provided tree sitter based mode ? I personally think this is the future, but not sure about others. :-)

Looking at how fast treesitter modes are merged into emacs, it certainly looks like it. It also will be a lot easier to maintain progmodes. I'm looking forward to it because rust-mode can be really annoying. Maybe the reason why I've been the only rust-mode maintainer for quite some time though there are lots of emacs users using rust.

I think we don't need a new rustic-mode. It should be possible to let the user choose if rustic should derive from rust-mode or the new rust-ts-mode. Shouldn't be too difficult, unless I'm missing something.

> Will this new mode will be on-par with the current existing regex based one ? If not, where do they differ ?

I have no idea what treesitter is capable of and how it compares to what rust-mode currently offers. But even if we have to use some kind of hybrid implementation, it will be a huge benefit. Indentation and font-locking alone can be a huge pain.

brotzeit:
I'm also curious how much overlap there is between treesitter and lsp. Seems lsp gradually indents to handle everything like systemd =)

psibi:
> Maybe the reason why I've been the only rust-mode maintainer for quite some time though there are lots of emacs users using rust.

If it helps, I can additionally volunteer to maintain some parts of rust-mode. I did initially use `rust-mode`, but once I discovered `rustic` - I never reached back to it. But now that these both aren't even a fork, I wonder if it makes sense to merge both of them.

>  It should be possible to let the user choose if rustic should derive from rust-mode or the new rust-ts-mode.

Sounds like a good idea and I would be to switch to the ts based one.

> I have no idea what treesitter is capable of and how it compares to what rust-mode currently offers.

I don't have a deep idea too, but from what I have seen so far - it looks to solve lots of hard problems in a much better way.

brotzeit:
[...]

PR #480, opened 2022-12-26
provide alternative rustic mode that derives from rust-ts-mode

brotzeit:
#475

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
     - [Installation](#installation)
         - [package](#package)
         - [straight](#straight)
+    - [treesitter](#treesitter)
     - [remote](#remote)
     - [Compilation](#compilation)
         - [Navigating errors](#navigating-errors)
--- a/README.md
+++ b/README.md
@@ -136,6 +136,29 @@ additional [installation
 instructions](https://github.crookster.org/switching-to-straight.el-from-emacs-26-builtin-package.el/)
 for moving your package management from package.el to straight.
 
+## treesitter
+
+You can try the new native treesitter mode `rust-ts-mode` with:
+
+```elisp
+(use-package rustic
+  :init
+  (setq rustic-treesitter-derive t))
+```
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -136,6 +136,29 @@ additional [installation
 instructions](https://github.crookster.org/switching-to-straight.el-from-emacs-26-builtin-package.el/)
 for moving your package management from package.el to straight.
 
+## treesitter
+
+You can try the new native treesitter mode `rust-ts-mode` with:
+
+```elisp
+(use-package rustic
+  :init
```
psibi, line 145: Probably this is more idiomatic:

```
:custom
(rustic-treesitter-derive t)
```

psibi (commented):
This looks great!

README.md
brotzeit: We need to set the variable before rustic.el is required. Unfortunately we can't set the parent mode dynamically(at least as far I know) so we have to configure this at initialization.

psibi: Ah, I see. Thanks!

brotzeit:
Ok let's give it a try. Shouldn't break anything, hopefully =)

brotzeit:
I'll revert it for now. Seems I only tested it with emacs master. I'll check again.

brotzeit:
The issue was probably that I tagged the variable with emacs29 afterwards...

CeleritasCelery:
Now that Emacs 29 is official released. Can this be merged again?

frtmelody:
@brotzeit Emacs 29 is official released, treesit should be supported?

dertuxmalwieder:
I'll +1 here. I would like to slowly migrate to `rust-ts-mode` (and others), but Rustic is a great help, so I would like those two to play together well.

psibi:
@liebkne and others: Can you give this a try and see how it goes: https://github.com/brotzeit/rustic/pull/527 ? I have been using it for a day and seems to be working fine.

PR #485, opened 2023-01-08
Fix org-babel workflow when auto wrapping main function

psibi:
This bug is easily reproducible when you use try something like this:

```
let x = [1,2,3,4,5,6];
println!("{x:?}");
```

And then try executing it. While you will get the result properly, you could see in the Messages buffer that that would be some errors related to compilation buffer.

And the reason is this: It tries to do rustfmt on the babel block and fails. In this fix, I just disable doing rustfmt when we have enabled automatic wrapping or when we instruct it to wrap the main function.

Another alternative fix could be do rustfmt on the wrapped main function and the remove the wrapped main function and render it again on the babel blocks. But IMO, that's going to be very fragile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ of rustic, however can be considered light-weight compared to some rustic's
 functionality.
 
 The shared functions and options exist as aliases in the rust-mode and
-rustic namespace for backwards compatability reasons(rustic has been a fork).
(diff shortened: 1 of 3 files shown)


brotzeit:
Works for me.

brotzeit (approved):

PR #486, opened 2023-01-08
Make envrc mode work

### LINKED ISSUE #420: Problem with local environments (`envrc`)
When using local environments with `envrc`, compilation fails because `cargo` is not found [1]. This is, because compilation is performed in a temporary buffer, and the local environment is not copied over. There has been a fix in `rust-mode` [2,3], but the relevant function `rust-buffer-project` is not used by `rustic`. Instead, `rustic-buffer-workspace` is used, which seems to be identical or at least very similar. Could we also implement the fix into the function used by `rustic`, or remove the duplicated code and use `rust-buffer-project`?

Thanks for your help!

References:
- [1] https://github.com/purcell/envrc/issues/12
- [2] https://github.com/rust-lang/rust-mode/issues/446
- [3] https://github.com/rust-lang/rust-mode/pull/447/files

tesujimath:
Propagate process-environment and exec-path to any auxiliary buffer which is created, so that [envrc](https://github.com/purcell/envrc) mode works.

An example requirement for this is with [Nix development shells via direnv](https://nixos.wiki/wiki/Development_environment_with_nix-shell).  With this, the Rust compiler, rust-analyzer, et al are not on the user's default path, but only on a local path configured by direnv within the project directory.

Two different mechanisms are used to make this work.

1. Code which makes use of `get-buffer-create` or `generate-new-buffer` is wrapped in [inheritenv](https://github.com/purcell/inheritenv)

2. A macro which wraps `with-temp-buffer` is used in place of that, since inheritenv does not handle that case, adapted from https://github.com/magit/magit/pull/4169

Subsequent additions of code which creates such auxiliary buffers should do likewise.

Additionally, LSP mode is started with `lsp-deferred` so that envrc has a chance to configure the path before it is required.

This change is working for me for cargo compilation and LSP mode rust-analyzer.  I am sure I'm not utilising the full surface area of rustic mode, so may have overlooked something.  Feedback welcome.

Fixes #420

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -67,25 +67,26 @@ considered."
 Compilation is started in directory DIR with appropriate
 TOOLCHAIN-KW-OR-STRING and MAIN-P indicates if a main function
 should be wrapped in which case we will disable rustfmt."
-  (let* ((err-buff (get-buffer-create rustic-babel-compilation-buffer-name))
-         (default-directory dir)
-         (toolchain (cond ((eq toolchain-kw-or-string 'nightly) "+nightly")
-                          ((eq toolchain-kw-or-string 'beta) "+beta")
-                          ((eq toolchain-kw-or-string 'stable) "+stable")
-                          (toolchain-kw-or-string (format "+%s" toolchain-kw-or-string))
-                          (t (format "+%s" rustic-babel-default-toolchain))))
-         (params (list "cargo" toolchain "build" "--quiet"))
-         (inhibit-read-only t))
-    (rustic-compilation-setup-buffer err-buff dir 'rustic-compilation-mode)
-    (when rustic-babel-display-compilation-buffer
-      (display-buffer err-buff))
-    (rustic-make-process
-     :name rustic-babel-process-name
-     :buffer err-buff
-     :command params
-     :filter #'rustic-compilation-filter
-     :sentinel (lambda (proc output)
-                 (rustic-babel-build-sentinel toolchain proc output main-p)))))
+  (rustic--inheritenv
+   (let* ((err-buff (get-buffer-create rustic-babel-compilation-buffer-name))
+          (default-directory dir)
+          (toolchain (cond ((eq toolchain-kw-or-string 'nightly) "+nightly")
+                           ((eq toolchain-kw-or-string 'beta) "+beta")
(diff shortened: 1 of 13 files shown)


psibi:
> An example requirement for this is with [Nix development shells via direnv](https://nixos.wiki/wiki/Development_environment_with_nix-shell).

I use pretty much the same workflow and have been using [envrc](https://github.com/purcell/envrc) for it which seems to well (at least for my workflows).

rustic-babel.el
```diff
@@ -9,6 +9,8 @@
 (require 'ob-ref)
 (require 'ob-core)
 
+(require 'inheritenv)
```
brotzeit, line 12: I'm afraid we can't just assume the package is present. Do you think we can use this functionality only when inheritenv is present ? I'm hesitant to make this an additional dependency, it seems it isn't used this way by many packages https://melpa.org/#/inheritenv

tesujimath: @brotzeit That's a good point!  I pushed up a new commit which only uses inheritenv if available.  How does this seem to you now?

tesujimath:
Squashed before rebase, sorry, as rebase unsquashed is just :weary:

rustic.el
```diff
@@ -163,6 +170,22 @@ this variable."
...
+;; with inheritenv.  Rather than depend on that package, we conditionally
+;; wrap if it is installed.  Users of envrc ought to ensure the inheritenv
+;; package is available before loading rustic.
+
+(defmacro rustic--inheritenv (&rest body)
+  "Wrap BODY so that the environment it sees will match the current value.
+This is useful if BODY creates a temp buffer, because that will
+not inherit any buffer-local values of variables `exec-path' and
+`process-environment'."
+  `(if (featurep 'inheritenv)
```
brotzeit, line 185: Do you think it makes sense to change it to `(or (featurep 'inheritenv) (require 'inheritenv nil t))` ? Just in case.

tesujimath: That doesn't make sense to me.

In order to activate this, prior to loading rustic we must already have done `(require-package 'inheritenv)`, after which `(featurep 'inheritenv)` will be true.

What scenario are you thinking of?

brotzeit:
Thanks!

tesujimath:
Thanks for merging this promptly!

PR #491, opened 2023-01-30
Remove dead code in `cargo-outdated`

mishazharov:
Previously I refactored cargo-outdated so that is does not use `rustic-cargo-outdated--skip-to-packages`. However, I forgot to actually remove the function definition. This commit does that

Sorry for all of the PR's, hopefully this will be the last one in a while


brotzeit:
No problem, thanks for the fix.

PR #499, opened 2023-03-26
fix(rustic-babel): disable toolchain when invalid/unneeded

yuuyins:
When the user installs Rust tools using a method other than rustup, e.g. using an operating system's package manager, cargo generally has no support for toolchain specification. In such case, the user can then `nil', or `""', so that the rescpective functions in `rustic-babel' will remove the toolchain from params, that is only toolchain has a valid value it uses toolchain.

See also: https://github.com/brotzeit/rustic/pull/279#issuecomment-1336164427


yuuyins:
i did this fix a while ago and works for me, unsure if it's fit. can't remember why i didn't PR

rustic-babel.el
```diff
@@ -73,9 +73,18 @@ should be wrapped in which case we will disable rustfmt."
...
                            ((eq toolchain-kw-or-string 'stable) "+stable")
+                           ((or (null rustic-babel-default-toolchain)
+                                (null (string-blank-p rustic-babel-default-toolchain)))
+                            nil)
                            (toolchain-kw-or-string (format "+%s" toolchain-kw-or-string))
                            (t (format "+%s" rustic-babel-default-toolchain))))
-          (params (list "cargo" toolchain "build" "--quiet"))
+          (params (remove nil (list "cargo"
+                                    (unless (or (null toolchain)
+                                                (= 0 (string-blank-p toolchain))
```
psibi, line 83: I belive this will be a problem if `toolchain-kw-or-string` is something like `1.64.0` since toolchain will be `+1.64.0` and evaluating something like this:

```
(= 0 (string-blank-p "+1.64.0"))
```

will result in error:

```
Wrong type argument: number-or-marker-p
```

rustic-babel.el
```diff
@@ -73,9 +73,18 @@ should be wrapped in which case we will disable rustfmt."
...
+                                                (equal "+" toolchain))
```
psibi, line 84: Any reason why we are doing this check ?

rustic-babel.el
```diff
@@ -126,7 +135,13 @@ execution with rustfmt."
 
            ;; run project
            (let* ((err-buff (get-buffer-create rustic-babel-compilation-buffer-name))
-                  (params (list "cargo" toolchain "run" "--quiet"))
+                  (params (remove nil (list "cargo"
+                                            (unless (or (null toolchain)
+                                                        (= 0 (string-blank-p toolchain))
```
psibi, line 140: Similar comments as above. I wonder if we can refactor it to a single function to simplify it.

psibi (changes requested):
Apart from some minor changes/questions, I think having an additional test would be good too.

rustic-babel.el
yuuyins: e removed that because that was redundant unnecessarily

rustic-babel.el
yuuyins: the + is in for just in case whatever it might be that no string was given to toolchain

yuuyins:
working  for me

yuuyins:
forgot about the test i'll have to look into it

psibi (commented):
Thanks, this LGTM. I'm willing to merge this without a test since the updated code is much easier to read, if you are okay with it. :-)

yuuyins:
@psibi yes, please you can merge it

psibi:
@yuuyins Sorry, it seems there is some test failures. Can you look into it ?

rustic-babel.el
```diff
@@ -73,9 +85,12 @@ should be wrapped in which case we will disable rustfmt."
           (toolchain (cond ((eq toolchain-kw-or-string 'nightly) "+nightly")
                            ((eq toolchain-kw-or-string 'beta) "+beta")
                            ((eq toolchain-kw-or-string 'stable) "+stable")
+                           ((or (null rustic-babel-default-toolchain)
+                                (null (string-blank-p rustic-babel-default-toolchain)))
+                            nil)
```
gcentauri, line 90: this bit doesn't look quite right to me... `string-blank-p` will return `0` if its an empty string, so `null` will be false.

I'm not sure if this is related to the failing tests, but if `rustic-babel-default-toolchain` _were_ set to `""` in the test environment, we wouldn't hit `nil` here and would fall through.

yuuyins: eq 0 now, thank you so much! if you have resources/time to see what's going on with the tests, that would be wonderful as I'm having to max focus on other stuff rn.

gcentauri: i will give it a shot, i need to set up all the dependencies locally to run tests, so it may be a bit before i can get to it.

StrictlyMonad:
@psibi mind approving the workflow for us to determine if the PR still fails or not?

PR #525, opened 2023-08-12
Use RUSTIC-CARGO-BIN in org babel code

easye:
If the "cargo" executable is not in the PATH, the org babel code will not work.  This was the case for me on macOS using rustup under Emacs.app, where the per-user PATH settings are not easily configured.


psibi (approved):
Thanks, LGTM. Unfortunately CI seems to be failing for a different reason that needs to be resolved before merging this.

Issue #509, opened 2023-05-23
Directory that binary executed from


barslmn:
Hi,
System Info:
Debian sid
Doomemacs
rustic version: 3.4

I am using rustic from orgmode by compiling rust code blocks.  I have the following files:
```
❯ ls -1
aoc2022.org
input1
```

And I have the following code block in the org file:
```
#+BEGIN_SRC rust :results output
use std::fs;

fn main() {
    let file_path = "input1";
    let contents = fs::read_to_string(file_path)
        .expect("Should have been able to read the file");
}
#+END_SRC

#+RESULTS:
: thread 'main' panicked at 'Should have been able to read the file: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/main.rs:7:10

```

Executing this block from org file gives the no such file error. In order to investigate I run the following block:

```
#+BEGIN_SRC rust 
use std::env;
fn main() {
    println!("{:?}", env::current_dir());
}
#+END_SRC

#+RESULTS:
: Ok("/tmp/babel-C1Nmcb/cargozPIbXS")
```

My question is how do I change the directory that binary is executed from /tmp/babel-XXXXXX/cargoXXXXXX to where the org file is? This does not happen with other languages (I tried C, python, shell) that's why I think rustic repo is the right place ask.
Thanks, B

ed9w2in6:
I face this issue too, which is a bug in my opinion as most if not all other babel executors all honours executing the binary in the directory specified by buffer local variable `default-directory`.

AFAIK, the current flow of source block execution is:
1. `rustic` ultimately uses `rustic-babel-eval` to build the compiled binary for the src block.
2. `rustic-bable-eval` uses `rustic--inheritenv` which does tries to retain some environment settings.
3. when constructing command for `make-process`, `rustic-babel-build-sentinel` is used for `:sentinel` later async execution 
4. Finally, and other `make-process` call will use `'("cargo" toolchain "run" "--quiet")` as command, in which toolchain is specified by `:toolchain`

From 2, we can see that this project does aim to preserve local environment when executing the compiled binary with `org-babel` framework.  So I hope we can agree that we should fix this issue.

Here, the culprit is at step 4.  By default, `cargo` will search for manifest file (i.e. `Cargo.toml`) from current directory or it parents, then execute the corresponding binary in the current directory.  This implies that one cannot run the binary via cargo outside of the workspace with introducing extra command line options.

IMO, the correct high level logic is:
1. somehow change directory to the original `default-directory`
[...]

PR #535, opened 2023-10-13
fix(rustic-babel-build-sentinel): align behaviour with other lang

ed9w2in6:
In an org file, the src block's code should be interpreted, OR the binary compiled from it SHOULD execute with PWD as `default-directory` of the org file.

Summary of changes: 
1. save the original default-directory then later restores it.
2. supply `--manifest-path` option to `cargo run` to execute the target `Cargo.toml` binary compiled from the src block.

For reference:
Try to execute a C++, C, js (node), python block.
In particular, see implementation of [org-babel-C-execute](https://github.com/emacs-mirror/emacs/blob/e58b1d371fbbb3d9a7a1898fce2d35766493aab0/lisp/org/ob-C.el#L131C17-L131C17) or [org-babel-python-evaluate-external-process](https://github.com/emacs-mirror/emacs/blob/e58b1d371fbbb3d9a7a1898fce2d35766493aab0/lisp/org/ob-python.el#L327).
May be we can re-write like them in the future.

fixes: https://github.com/brotzeit/rustic/issues/509

### DIFF
--- a/rustic-babel.el
+++ b/rustic-babel.el
@@ -116,7 +116,8 @@ execution with rustfmt."
    (let ((proc-buffer (process-buffer proc))
          (inhibit-read-only t))
      (if (zerop (process-exit-status proc))
-         (let* ((default-directory rustic-babel-dir))
+         (let* ((original-default-directory default-directory)
+                (default-directory rustic-babel-dir))
 
            ;; format babel block
            (when (and rustic-babel-format-src-block (not rustic-babel-auto-wrap-main) (not main-p))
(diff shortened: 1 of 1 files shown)


ed9w2in6:
The 2 code block should yield identical outputs when executed in a src block of a org file.

## Rust
```rust
use std::env;
use std::fs;
use std::io;

fn main() -> io::Result<()> {
    for f in fs::read_dir(".")? {
        let fpath = f?.path();
        let fname = fpath.to_str().expect("Path should convert to str");
        println!("{fname}");
    }
    Ok(())
}
```

## C++
```C++
#include <string.h>
#include <dirent.h>
#include <stdio.h>

using namespace std;

int main() {
DIR *dir;
struct dirent *dp;

dir = opendir(".");
while ((dp = readdir(dir)) != NULL) {
  // printf("debug: %s\n", dp->d_name);
  if ( !strcmp(dp->d_name, ".") || !strcmp(dp->d_name, "..") ) {
    // do nothing (straight logic)
  } else {
    printf("./%s\n", dp->d_name);
  }
}
closedir(dir);
return 0;
return 0;
}
```

Here is a sample org file for your reference.
<details>
  <summary>Sample Org File</summary>

```org
#+TITLE:Loop Do Nothing
* Components
** foobar
   #+begin_src rustic :results output scalar :eval yes :main no
use std::env;
use std::fs;
use std::io;

fn main() -> io::Result<()> {
    for f in fs::read_dir(".")? {
        let fpath = f?.path();
        let fname = fpath.to_str().expect("Path should convert to str");
        println!("{fname}");
    }
    Ok(())
}
   #+end_src

#+begin_src C++ :results scalar :eval yes :includes <string.h> <dirent.h> <stdio.h> :namespaces std
DIR *dir;
struct dirent *dp;

dir = opendir(".");
while ((dp = readdir(dir)) != NULL) {
  // printf("debug: %s\n", dp->d_name);
  if ( !strcmp(dp->d_name, ".") || !strcmp(dp->d_name, "..") ) {
    // do nothing (straight logic)
  } else {
    printf("./%s\n", dp->d_name);
  }
}
closedir(dir);
return 0;
#+end_src
```

</details>

PR #536, opened 2023-10-17
syntax error in `rustic-cargo-add` with prefix argument

sigvei:
`rustic-cargo-add` with a prefix command called `read-from-minibuffer` with erroneous arguments.

The current master produces this error message:

`funcall-interactively: Wrong type argument: keymapp, " add "`

This is due to the line `(read-from-minibuffer "Cargo add command: " (rustic-cargo-bin) " add ")`, which makes " add "  a third argument to `read-from-minibuffer`. I guess it's suppose to just be concated with `rustic-cargo-bin`.

This commit implements that and also DRY's the code by storing the base command in a let binding.

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -744,11 +744,10 @@ The documentation is built if necessary."
   "Add crate to Cargo.toml using 'cargo add'.
 If running with prefix command `C-u', read whole command from minibuffer."
   (interactive "P")
-  (let* ((command (if arg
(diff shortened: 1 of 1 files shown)


LionyxML:
@brotzeit  this is very simple and very useful, do you think you could review it in a near future?

Thanks @sigvei

psibi:
@LionyxML Unfortunately I have been not able to reach @brotzeit. I maintaining a temporary (hopefully) fork here: https://github.com/psibi/rustic

PR #539, opened 2023-12-05
Support goto-error like 'compile' for 'clippy'

roife:
Currently, rustic only supports `goto-error` in `rustic-compilation-mode`. 

image

However, the output format of Clippy is the same as rustc. Thus, the parser in `rustic-compile-goto-error-hook` can be reused for `rustic-cargo-clippy-mode`


roife:
Oh I find it is better to use `derived-mode-p`, so `rustfmt` and `cargo check` will also benefit from it.

PR #551, opened 2024-02-19
Revamp CI, Enable it for 29.1, Fix flaky tests

psibi:
The changelog has more details.


psibi:
@samhedin Any chance you can review this and approve ? I think @brotzeit has been out of reach for some time now.

psibi:
Fair enough! I would like to avoid maintaining my own fork if possible, but I don't think it's possible anymore.

PR #565, opened 2024-05-14
Add unique test module scope for arg to 'cargo test'

cebernard:
When rustic-cargo-current-test sees point at a test module's definition (or slightly below, while being above the first test function), then construct an argument for 'cargo test' that will run only that module's tests.

Since most test modules are not uniquely named, this means either prepending the parent module to the current module (e.g. when tests are in 'some_mod.rs', then 'cargo test some_mod::tests'), or if at crate root then using 'lib' or 'main' as the argument (e.g. when tests are in 'lib.rs', then 'cargo test lib').

Handle both module definition styles, either when modules are defined in a like-named file or living in a like-named directory in a file named 'mod.rs'.

This commit is feature complete, but only covers the scenario where tests are defined in the same file as production code.

Partially resolves brotzeit#561 (1-3).

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -199,7 +199,12 @@ When calling this function from `rustic-popup-mode', always use the value of
 
 ;;;###autoload
 (defun rustic-cargo-current-test ()
-  "Run 'cargo test' for the test near point."
+  "Run 'cargo test' for the test or module near point.
+When point is at or below the line defining a test function, run
+that test.
+
+When point is above all test functions and at or below a test
+module's definition, then run that module's tests."
   (interactive)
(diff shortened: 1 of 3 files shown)


cebernard:
Hi everyone. Just checking in to see if there's any feedback on this PR. I'm happy to make any adjustments based on your suggestions.

psibi:
This repo is unmaintained, this is the new home: https://github.com/emacs-rustic/rustic

More details here: https://psibi.in/posts/rustic.html

cebernard:
Thanks. Somehow I missed the memo. I guess I'll duplicate this PR on that repo.

PR #574, opened 2024-07-26
Add `rustic-test-failure` to `compilation-error-regexp-alist`

### LINKED ISSUE #573: compilation-error-regexp does not always match test outputs
I'm not totally sure when this changed, but the `compilation-error-regexp` defined by rustic no longer matches the the output when a test assertion fails.

My test output generally looks like this:

```
thread 'module::submodule::subsubmodule' panicked at lib/rs/tests/test-module/models/db-model.rs:454:5:
```

The value added for `rustic-panic` in `compilation-error-regexp-alist-alist` is currently:

```elisp
"thread '[^']+' panicked at '[^']+', \\([^\n]+\\):\\([0-9]+\\):\\([0-9]+\\)"
```

The inclusion of the `'[^']+', after "panicked at" is what seems to prevent the match.

I have worked around it locally by adding the following to my `use-package` config for `rustic`:

```elisp
(use-package rustic
  :hook
  :hook
  (rustic-cargo-test-mode
   . (lambda () (add-to-list 'compilation-error-regexp-alist
                             `(,(rx "thread '"
                                    (one-or-more (not "'"))
                                    "' panicked at "
                                    (group (one-or-more not-newline))
                                    ":"
                                    (group (one-or-more digit))
                                    ":"
                                    (group (one-or-more digit)))
                               1 2 3))))
```

which generates the regexp `"thread '[^']+' panicked at \\(.+\\):\\([[:digit:]]+\\):\\([[:digit:]]+\\)"`.

With this in place, jumping back and forth between tests in the rustic test output compilation buffer works again as expected, along with jumping directly to the failing test.

neosimsim:
Add regexp to `compilation-error-regexp-alist` to match test failures for `rustic-cargo-test-mode`.
This make test output like 

```
---- fail stdout ----
thread 'fail' panicked at src/main.rs:20:5:
assertion `left == right` failed
  left: 1
 right: 2
```

navigatable.

Closes #573

### DIFF
--- a/rustic-cargo.el
+++ b/rustic-cargo.el
@@ -143,8 +143,22 @@ stored in this variable.")
     map)
   "Local keymap for `rustic-cargo-test-mode' buffers.")
 
+(defvar rustic-test-failure
+  (let ((test "---- [^\n]+ ----\nthread [^\n]+ panicked at ")
+        (file "\\([^\n]+\\)")
+        (start-line "\\([0-9]+\\)")
+        (start-col  "\\([0-9]+\\)"))
+    (let ((re (concat test file ":" start-line ":" start-col)))
+      (cons re '(1 2 3))))
+  "Match test failures.")
+
 (define-derived-mode rustic-cargo-test-mode rustic-compilation-mode "cargo-test"
   :group 'rustic
+
+  (add-to-list 'compilation-error-regexp-alist-alist
+               (cons 'rustic-test-failure rustic-test-failure))
+
+  (add-to-list 'compilation-error-regexp-alist 'rustic-test-failure)
 
   (when rustic-cargo-test-disable-warnings
     (setq-local rustic-compile-rustflags (concat rustic-compile-rustflags " -Awarnings"))))


neosimsim:
Please note that I'm not so sure about the error type / level. Mabe 0 or 1 would be more suitable.

rustic-cargo.el
```diff
@@ -143,9 +143,23 @@ stored in this variable.")
     map)
   "Local keymap for `rustic-cargo-test-mode' buffers.")
 
+(defvar rustic-test-failure
+  (let ((test "---- [^\n]+ ----\nthread [^\n]+ panicked at ")
```
mplanchard, line 147: I'm not sure about including the initial `---- blah ---` bit, because tests that print to stdout will have their output below that but before the panic, e.g.:

```
---- entity::models::entity::insert_or_update_all_entity_types stdout ----
[lib/rs/spec-test/src/lib.rs:31:9] "TEST_ENVIRONMENT already set" = "TEST_ENVIRONMENT already set"
thread 'module::models::some_model::insert' panicked at lib/rs/lib-data/tests/lib-data-test/entity/models/entity.rs:89:53:
called `Result::unwrap()` on an `Err` value: EntityDbError(SqlxStoreError(Unexpected(Unexpected("failed to deserialize model: invalid type: null, expected string or map"))))
```

neosimsim: Good point. But I'm not so sure that my approach is correct, anyway. I believe you're right in #573 that `rustic-compilation-panic` is not correct.

neosimsim:
The more I think about it, the more I feel this was the wrong approach and `rustic-compilation-panic` needs to be fixed. #575

neosimsim:
Closed in favor of #575.

Issue #566, opened 2024-05-31
Maintained fork created at emacs-rustic/rustic


CeleritasCelery:
A maintained fork of this repo has been created at https://github.com/emacs-rustic/rustic

wusticality:
> A maintained fork of this repo has been created at https://github.com/emacs-rustic/rustic

Is this repository dead then?

CeleritasCelery:
It is unmaintained. Meaning this repository will get no fixes or enhancements (at least if/when the owner returns).

PR #575, opened 2024-07-26
Fix `rustic-compilation-panic` regex

### LINKED ISSUE #573: compilation-error-regexp does not always match test outputs

neosimsim:
Fix `rustic-compilation-panic` regex by removing ` '[^']+',`

Closes #573


neosimsim:
I added the test in a separate commit, so the regression is easier. Unfortunately I cannot figure out why `compilation-num-errors-found` jumps from 0 to 2.

neosimsim:
> I added the test in a separate commit, so the regression is easier. Unfortunately I cannot figure out why `compilation-num-errors-found` jumps from 0 to 2.

I suspected `rustic-panic` to be in `compilation-error-regexp-alist` twice, but that not the case

```
(rustic-panic rustic-info rustic-warning rustic-error)
```

psibi:
Note about https://github.com/brotzeit/rustic/issues/566

neosimsim:
> Note about #566

Oh my goodness. I totally missed that. Thanks for pointing that out @psibi

neosimsim:
Closed in favor of emacs-rustic/rustic#31

PR #593, opened 2026-05-31
Fix byte-compilation warnings

dannywillems:
## Summary

This PR fixes byte-compilation warnings in rustic. Every change here is
behaviour-preserving; the rendered documentation, control flow, and
public API are unchanged. Each warning type is in its own commit.

This is an automated effort to help maintain packages in the Emacs
community.

## Fixed warning types

- **Docstrings with unescaped single quotes**: many docstrings wrapped
  shell commands and code fragments in lone single quotes (for example
  `'cargo test'`, `'clippy --fix'`, `'fn main'`, `'Cargo.toml'`). The
  byte-compiler reads a lone `'` as the start of an unbalanced quoted
  form. Escaped with `\='` (or `` `...' `` when naming an Emacs symbol).
  The help buffer still renders plain single quotes.
- **Obsolete `cl-gensym`**: replaced with `gensym`, available in core
  since Emacs 26 and identical in behaviour here.
- **Docstrings wider than 80 columns**: reflowed so the first line is a
  complete sentence under the limit and the remaining text wraps.
- **Unused lexical variables and arguments**: underscored arguments
  required by calling convention but unused in the body (process
  sentinel lambdas, an interactive optional argument, a `cl-defmethod`
  specializer) and dropped a dead `let` binding in
  `rustic-cargo-spellcheck-command`.
- **Obsolete `if-let` / `when-let`**: replaced with `if-let*` /
  `when-let*`, available since Emacs 26 with identical binding and
  short-circuit semantics. The single-variable shorthand in
  `rustic-cargo-test-dwim` is wrapped in the list-of-bindings form the
  starred macros require.

## CI

Adds a byte-compile + load GitHub Actions workflow (Emacs 28.1, 30.1,
snapshot) plus dependabot. It does not use warnings-as-errors because
the package still emits the out-of-scope warnings below. (rustic
declares Emacs 26.1 as its minimum, but the current MELPA dependency
set no longer installs on 26.1, so the lowest matrix entry is 28.1.)

## Out of scope (left as residual)

These are behaviour-changing or relate to deliberately optional
integrations, so they are intentionally not touched:

- `replace-buffer-contents` is obsolete as of 31.1; `replace-region-contents`
  has a different signature, so swapping it is behaviour-changing.
- `project-roots` is obsolete; `project-root` returns a single root
  rather than a list, so it is behaviour-changing.
- `defcustom rustic-popup-commands`: `list` without arguments is a
  `:type` widget change.
- `Unused lexical argument 'arg'` in `rustic-run-shell-command` is
  entangled with an `assignment to free variable 'command'` issue; the
  function appears half-implemented and fixing it would change behaviour.
- `value from call to ... is unused` (`length`, `f-exists?`,
  `file-exists-p`): discarded return values that look like real logic
  bugs; changing them alters behaviour.
- `global/dynamic var 'eglot' lacks a prefix`: references an external
  variable from the eglot package.
- `function ... is not known to be defined` for `inheritenv-apply`,
  `lsp-*`, `eglot-*`, `flycheck-*`, `flymake-*`, `define-*mode`
  (polymode), `jsonrpc-request`: all optional integrations that are not
  declared dependencies and are loaded conditionally.
- `assignment to free variable` for `rust-*` and `flymake--*` and
  `function might not be defined at runtime` for forward references:
  load-order / dynamic-binding concerns that are out of scope.


dannywillems:
This is an automated effort to help maintaining packages in the Emacs community. See https://x.com/dwillems42/status/2060720730338185699 and https://dannywillems.github.io/emacs-package-maintenance/

dannywillems:
Closing: this repo appears dormant. The active rustic is emacs-rustic/rustic, where I have opened the equivalent fix: https://github.com/emacs-rustic/rustic/pull/128

